PR #3, opened 2015-05-29
0.5.0.10E

Ealb:
- Completed muuch Atlantians
- Updated armor and weapon .csv to include recent items that NG currently uses
- release candidate; generated new executable
- last external release for the 0.5.0.xxE ? Should we formally merge back to 0.5.1 for the next release and go ahead from there?


elmokki:
From my point of view you could've used normal version numbers right away when you started working on this. Your choice. Even 0.6.0 would not be completely outlandish version number considering that I've changed the minor number for less.

I was gonna work on some GUI stuff and possibly national spells, but I got sidetracked a bit.

Ealb:
I'll go ahead and make the next release 0.6.0 - if I twist my arm, I'll admit that the last 8-10mo of dev probably merit it, even if the engine hasn't changed much.

There's a few bugs I want/need to squash with monsters (the issue with hydras seems less resolved than I'd thought, and custom graphics still need done).

After that, my temptation would just be to plow on with content, but that would be easier/cleaner if with a couple more tools. The ones that seem the most useful that are lacking would be themes, just because that would essentially unlock a huge amount of variations among nations w/o needing to add new content, and some sort of by-nation "invariant/weighted slot(s)" chance to allow for all units with access to a slot of that sort to either pick it by default for that particular nation, or have it more heavily weighted. I.e., sub-races, mostly. E.g., every unit in a particular Caelian nation picks/leans toward picking raptor_wings for their overlay slot, or every muuch picks/leans towards red/white/brown/green basesprites, with associated stat variation. Etc. It's nothing we haven't discussed before, nor anything we can't technically do now with additional nations, but both of these would help fight redundancy and bloating with the content. E.g., the four slaver racial variants I added in .09E are interesting variations, but they really should be handled as themes rather than four separate races.

In the short term, I'll probably pick away at content (leaning towards trolls, ogres/giants, or bakemono (the latter was what I'd intended to do before I, uh, sorta accidentally made vaettir instead)) until I get too frustrated by some random shortcoming in the engine, and then dive back into the code. Not the ideal development strategy, but it ensures I'll be engaged and enthusiastic when I'm coding...

Ofc, when the next Dom4 patch comes, I'm pretty sure I'll have a pile of data revision needing done, especially with more explicit definitions of bronze/meteoric iron weapons coming in light of the rust change. It's perhaps also worth considering if we might want to look at revisiting the togglable-UW-nations issue, though I'd prefer that to be preceded by themes so we can still have the distinctly land-dwelling iron-forging "refined" Atlantians who have turned their backs on the sea...

elmokki:
The themes are easy. I can probably make them in a week or so if I don't run to a very high lack fo motivation. 

Locking a nation to certain item in certain slot may require some work considering how mages and troops etc are separate poses, but it's not really hard. It's just that the simplest solution may be a bit of a spaghetti one. I'd probably go for slot + item name and some command. It's easier for something like wings than for weapons at least. This is something that could really warrant a rewrite in general unit equipment generation code and unifying it for troops and mages.

PR #7, opened 2015-06-01
Minotaurs, graphics, filters

Ealb:
- 1/4 - 1/3 done with minotaurs (yes, I know, they weren't anywhere near the list of "things I'm going to work on")
- Some graphics fixes, mostly minor (stray pixels, realignments, etc)
- Some filter fixes: Fomorian mage survivals, fireshield graphics, and set heretic/slimer to +1 and fear to +5 (Dom4 minimums)

* Are mage filters supposed to be able to circumvent "#notfortier X" via tier variation? I've seen a number of tier-3-only filters showing up on tier-2 mages when the tier-3s have it as well; in the case of battlesummons, that seems pretty unbalanced even by NG's forgiving standards.


elmokki:
All content is good content. Well, as long as it's quality content, but you don't have issues with making such.

The mage filters are definitely a bug. I was thinking of rewriting them a few days ago anyway (but yet I haven't even touched themes yet). Currently, as you probably know, it's x filters given to mages and then they are applied to all mages whenever possible. I think that a better solution would be to roll filters for each tier separately with high chance of lower tiers getting the same filter as higher tiers if possible. This way, as a bonus, we wouldn't always have stuff like 3 primary mages, one of which has astral magic and thus ethereal, but the others having nothing cool. That is okay, but instead of always having it like that, it would be cool if say, second variant had earth magic and invulnerability and the third variant had fire magic and fire shield. It'd also allow removing some of the forging filters which at least tried to give forging only to secondary mages and things like that.

Ealb:
It's something I've thought about. My dev planning cycle unfortunately tends to include a phase that goes something like "get an idea for something; go off and start combing through the sprites for raw materials; see something that looks even cooler; drop everything and work on that instead", and some of the special weapons have definitely pushed me towards dropping everything. I may look into it at some point here.

Something that I've thought about as well would be adding poses with an explicit "hero" (and/or "multihero") role that could only be chosen for heroes, so as to allow for more impressive chassis and gear. I was mostly thinking nephilim at the time, but it could work even just with more "shiny" gearset variants of normal units for some races, although with that it might just be better to give various poses a graphically-appropriate fancy item file or two with the appropriately-exclusive tags.

Ealb:
On a somewhat idle note related to that last bit, there's also a part of me that would like to see #prophetshapes added for some commanders, but as that would pretty much require elite versions of sprites being drawn/defined for elite/tier-3 items, it'd probably be more far work than it would be worth.

elmokki:
Adding a hero pose would be very easy in terms of code.

Also regarding #prophetshapes, attack sprites in general would be a very similar thing in terms of code and graphics.

PR #9, opened 2015-06-03
Theme pull

elmokki:
Themes added

### DIFF
--- a/data/filters/default_priestfilters.txt
+++ b/data/filters/default_priestfilters.txt
@@ -2,9 +2,10 @@
 #name "awe"
 #basechance 0.1
 #command "#awe +1"
-#chanceinc "magic astral 0.05"
-#chanceinc "magic death -0.1"
-#chanceinc "magic blood -0.1"
+#chanceinc "personalmagic astral 0.5"
+#chanceinc "personalmagic death -0.1"
+#chanceinc "personalmagic blood -0.1"
+#power 2
 #end
 
 #new magefilter
@@ -13,7 +14,7 @@
 #command "#stealthy +10"
 #command "#gcost +20"
 #tag "notfortier 3"
-#tag "notfortier 2"
+#tag "notfortier 2" 
 #end
 
 
@@ -39,10 +40,11 @@
 
 #new magefilter
 #name "spy"
-#basechance 0.1
+#basechance 0.25
 #command "#stealthy +10"
 #command "#spy"
 #command "#gcost +20"
+#power 2
 #tag "notfortier 3"
 #tag "notfortier 2"
 #end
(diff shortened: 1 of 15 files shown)


elmokki:
Oh yeah, also, *0 is not a good idea for default filters in any case since that might make some nation unable to generate since it's all primitive or something like that. For more drastic stuff the #nationthemes command and a custom theme set is preferable.

Ealb:
This sounds really good. I agree that it's better to leave rare random chances for variety's sake instead of making cohesion the be-all end-all.

So right now themes just act as chanceincs? Hmm. Would it be possible for themes to have nation commands associated with them?  Some of the national commands like role chanceincs could be offloaded into themes on poses (e.g., Abysian ranged units are tagged "advanced"), but others would work better if we could have themes add nation commands. E.g., if it's "advanced" there's "#magicpriority earth 10" "#magicpriority nature 0.5"? Or "ice" Atlantians getting priority for water but "coral" getting a mix of water and nature, and iron getting mostly earth? Or "primal" Abysians getting #spreadheat +1" and #nationcommand "#idealcold -3", but "advanced" ones only getting ""#idealcold -2" and no heat? That sort of thing? If the option for aquatic nations gets added, that would be prime fodder for this. Oh, and also if we can add "#secondaryracecommand" I can get rid of my four stopgap slaver nations and just make a slaver theme. (Hmm, it would probably also be nice if primary races could modify their chances of various secondary races via themes - thus, subterranean Abysians have subterranean allies, slavers pick on weaker races more, monsters are more out of style when you become "advanced"...)

If nothing else the fact that this is available means I'm going to have to rework minotaurs, as I'd currently done "civilized +att/def" poses and "savage +berserk" poses, which could both occur simultaneously but were weighted according to era; now I'll probably just make the two have different bases and tie them to themes. Heh. Overhauling the nations to fully integrate themes is going to be a lot of work, but it'll be interesting...

elmokki:
Yes, currently they just inject chanceincs to every chanceinc decision there is.

Current functionality doesn't allow things like you wrote, but honestly it would be fairly trivial to just add new variables to the Theme class and read all the themes of a nation when needed. 

It would actually be pretty trivial to make themes inject certain lines to race definitions and that could already do all except the role chanceincs and something like dual wielding probability.It might need some changes to race command handling though. As for roles I think I could just make the role selection filter based fairly easily. That way you could just literally tag whole roles with theme tags and you could probably use some other chanceincs as well (though troop generation is pretty early so something like "more ranged for nations with air mages" is not possible). 

As for the part about secondary races, that's trivial. I'll move the theme selection right after the primary race selection and then it's either adding the races a chanceinc "hastheme" or adding a theme chanceinc based on race name. I think theme related things (as in the theme rulesests, not #theme commands) should generally be implemented in themes. A central place for adjusting what a certain thing does is nice and something like #chanceinc theme subterranean *10 is a bit cryptic compared to having #themeinc race agarthan *10 in the subterranean abysian theme file.

Ealb:
For debugging purposes, could a comment be added to the .dm somewhere about what themes were applied?

PR #11, opened 2015-06-03
Elmokki

elmokki:


elmokki:
Helpful note in addition to the details in the commit: #type <something> in themes as well as most filters (I think?) should prevent two of the same type being ever generated. I added #type techlevel to primitive and advanced to avoid primitive advanced nations.

Ealb:
I'm going to screw around trying to thoroughly theme Amazons tonight - they have a very good selection of bronze and normal armor, even if their weapons are all default. They're also a perfect candidate for a slave-keeping society. So I should have plenty of room to put some different types of themes through their paces.

Ealb:
Hmm, there seems to be a problem with #racedefinition in terms of #unitcommand (and possibly other things, but definitely that). It looks like the arguments to #unitcommand are getting completely tokenized, so that if I give e.g.[[ #racedefinition "#unitcommand '#prec 13'" ]] I'll end up with the line [[ "#prec ]] in the .dm file (whereby in [[X]], X is the actual text input or output from/to the theme or .dm file). While this seemed like the most intuitive syntax, I tried every variation I could reasonably conceive of that command in terms of quotes and substrings (up to and including escape characters), and the results varied from crashes to nothing at all being output in the .dm. [[#racedefinition "#unitcommand #flying"]] works fine. Entries like [[#racedefinition "#magicpriority astral 10"]] appear to be working fine as well. Digging into the code, in addOwnLine, str is passed in as " #unitcommand '#prec 13'", and parseArgs is tokenizing it to [#unitcommand, '#prec, 13'] rather than  [#unitcommand, #prec 13] as would be hoped.

PR #17, opened 2015-06-04
Elmokki

elmokki:


elmokki:
I think we're very close to themes being able to generate basic uw nations. I think all we need is a chanceinc to set if a setting disables uw nations and probably an easy way to restrict non-amphibians as secondary race (it kinda is possible already by tagging them nonamphibious I guess)

However, I detest UW. I think that that setting will be off by default.

Ealb:
UW is kinda flat, yeah. It was my first love way back in Dom1, which is why I so dearly *want* to like it but it just hasn't changed as much as the rest of the game has since then. It remains to be seen if the looming UW patch can fundamentally alter that dynamic; I'm not optimistic, but it'll probably make it a bit better. I suppose the "iron rusting" thing will at least make UW a bit more than just "very samey, only harder to move and less terrain variety".

I'm wondering how much of a PITA it would be to do a nation where every unit has a landshape and seashape (mermen/mersatyrs/etc.); it mostly just seems like it'd be a dynamic secondshape that carries over equipment, and the equipment would need a secondshape-type tag - either an extra sprite/recolormask, or  #tag "#aquaticversion foo". The latter would probably be better so it's possible to give different characteristics on land and sea. Still, that's completely idle chatter at the moment...

PR #18, opened 2015-06-05
Amazon themes and Race.java fix

Ealb:


Ealb:
Okay, something very strange is going on with this, and I don't think my changes could have caused it, but who knows. In any case, line 60 if Race,java is going out of bounds on the second nation (but not the first!) when > 1 are generated unless I add (c.args != null && c.args.size() != 0 && (stuff)), but at that point, I can see that themes start stacking from one nation to the next. So if I make 10 nations in a mod and all have a theme giving #mor +1, the 1st will give +1 and the 10th will give +10. Ish. It doesn't seem 100% consistent on this. I'm not readily seeing why, but I'm also not able to keep looking tonight. I'll resume doing so sometime tomorrow assuming it's still a problem.

(Also, yow, what was I thinking making #slaves gcost *0.8? That's way too generous even with the stat penalties. -2 gcost it shall be...)

elmokki:
Yeah. I thought of that before I go to sleep. Nations need to get copies of races instead of races themselves or we need to do some really annoying workarounds. I'll go with copies.

Ealb:
I think the new fix may have broken some command argument sanitizing. I wasn't getting secondary race commands added to the .dm until I added the following to addOwnLine at line 100 of Race.java

if(args.get(j).startsWith("\'") || args.get(j).startsWith("\""))
	args.set(j, args.get(j).substring(1));
if(args.get(j).endsWith("\'") || args.get(j).endsWith("\""))
	args.set(j, args.get(j).substring(0, args.get(j).length()-1));

...and I'm still seeing names like #name "'Ranger' of Belhimthos" rather than #name "Ranger of Belhimthos"

elmokki:
Please commit or post some suitable themes and I'll look at it tomorrow
5.6.2015 19.22 Ealb <notifications@github.com> kirjoitti:I think the new fix may have broken some command argument sanitizing. I wasn't getting secondary race commands added to the .dm until I added the following to addOwnLine at line 100 of Race.java

if(args.get(j).startsWith("\'") || args.get(j).startsWith("\""))
    args.set(j, args.get(j).substring(1));
if(args.get(j).endsWith("\'") || args.get(j).endsWith("\""))
    args.set(j, args.get(j).substring(0, args.get(j).length()-1));

...and I'm still seeing names like #name "'Ranger' of Belhimthos" rather than #name "Ranger of Belhimthos"

Ealb:
The thing with names doesn't seem to be theme-related. It looks like it's coming up from how bonusranged.txt (for a number of races, but for basic humans first and foremost) tags item-based name variants: #tag "name infantry 'Ranger'" ... #tag "name infantry 'Axe Thrower'" ... etc. This just started including the ' ' with this last patch, and at the same time I was having the secondary race commands (but not the primary race commands) not parse until I inserted the above sanitizing code, which combined with the timing points to both problems having a single source. I didn't really see other problems, though, and successfully ran a mod I generated with the patch.

I'll commit revised Amazon themes tonight, but the problems as described should exist with the slaver theme in the version currently in the repository.

Also, are theme chance modifiers being taken into account choosing poses and/or bases? I tried making a slave and free version of the amazon auxiliaries, first with basesprites (which IIRC is what's currently in the repository) and then with poses, and neither seem to work.

elmokki:
It seems that the very logical change I made to tag handling killed 
those 's when I fixed the way tags are read from the file. You can 
probably fix this by simply getting rid of the 's in the definition now 
which is what your code snippet seems to do. I think that may be a 
better idea than coding around it since the 's are pretty illogical 
anyway. I'll look at it closer tomorrow. Right now I have a cake to bake 
and am slightly drunk from a faculty end of spring semester dinner.

I also realized before the dinner that the event system allows making a 
foul spawn nation just like I want to make one, ie recruiting categories 
instead of troops. Like, you recruit "ranged foul spawn" or "large foul 
spawn" and get one of the many options from the category. Implementing 
that to nationgen would be dumb, but I really want to make a normal mod 
about it so it may take some of my time in the near future. More 
summer/end of spring semester stuff coming in the next weeks too though.

Ealb:
No worries.

I think I found and killed it. I added an additional ' -> " conversion in Entity.handleOwnCommand, and between that and my previous code snippet all the quotation problems *seem* gone. Didn't figure out what was up with bases/poses not processing chance increases from theme tags, but I really dig very deep after rooting out the "'" problem and getting distracted implementing a basic version of the weapon-commander-name-variants (50% chance per generic commander to check if such a name is available)...

elmokki:
Yeah, the ' -> " conversion is a good enough solution. The reason I 
figured removing 's is that then you could use 's in names, but then 
again we're talking about the kind of names that are very unlikely to 
use those.

Poses aren't being handled by chanceinchandler in most places for some 
reason and even items not in all places. This is because the whole 
concept of chanceinchandler was born when I started working on filters. 
I fixed the majority of this, but I got to say, troopgenerator and 
sacredgenerator really call for a rewrite.

PR #23, opened 2015-06-07
 Vaetti themes

Ealb:


elmokki:
Would there be point in making something like racial #guaranteedtheme tribal to guarantee a tribal theme? I think we'd also need some probability commands so it can be chosen at what probability the nation should get even more themes. But I mean, it kinda seems like the tribal thing would work very well on top of the other stuff.

Also, once 0.6.0 is released I think we should stop merging to master. Let's just make a development fork and merge to it until 0.6.1.

elmokki:
Also, I'll see if I can rewrite unit generation somewhat at some point before 0.6.0. It might take a while before I get enough motivation to do it substantially enough though so might be better to release 0.6.0 once you feel like you're finished with all that wonderful content.

I think the base principle of unit templates is fine, but we also need far more common militia and generic elites. Those can be hard coded with a toggleable switch and the templates can just be more special units. I think some equipment should be tagged by militia (also same for scout. Priest might be a good tag as well, though priests can be separated to different poses already anyway) like it is by elite and sacred. I also think that instead of the current system where elite and sacred items are filtered out / everything else is filtered out is bad. Instead sacred/elite stuff should generally be filtered out for basic troops, but for elites instead I think there should be a temporary theme like chanceinc that just makes the probabilities tenfold or more (depending on how special look you want I suppose). 

Also the new system needs to be such that mage generation can have naked units with magic paths so gear can be chosen after paths are selected. After all, the antlered helmet for one really cries for nature magic and there are some other similar stuff as well. #varyslot for mages could use a probability so you can do something interesting like make mages vary mount for each variant at very low probability.

All in all it would be ideal to have a more standardized troop generation process that can be used for sacreds and troops (and mages) without too much hassle.

Ealb:
I could see the value of a #guarenteedtheme, yeah. If we were able to arbitrarily up the number of themes a nation could have, that could work quite nicely in place of the "per-nation fixed slot choice" idea I'd discussed before for subraces by just making e.g. Raptors or red/white/brown/green muuch or reef/ocean/deep Atlantians as tribal themes. Right now, I was writing tribal themes with the assumption that you wouldn't always have one (and thus charging a gcost premium to get one in the case of vaetti) but it wouldn't be at all unreasonable for some races to always have something like that - with vaetti wolf would pretty much be a natural default tribe, after all.

Making a dedicated dev fork sounds like a good idea, yeah.

As to when the next batch of content is "done", um, er... I normally just go until I feel guilty for having let so much pile up. I'd like to get a good chunk of theme work done to be released with the new functionality, and that actually makes me think that after I finish minotaurs it might be good to pick back up where I'd paused on barbarians since they'd do the tribal stuff more or less the same and would make a nice addition to showcase themes along with some of the others. That shouldn't be that involved unless I were to go overboard with mages, as barbarians will be pretty straightforward and narrow with their troops, graphically and mechanically speaking.

I agree emphatically with what you're saying about how elite, etc. gear is handled. I've been kludging around in that direction with some of the races I've done recently by making their elite gear untagged but very low probability with a fair-sized chanceinc for sacreds; changing the whole all-or-nothing all-non-elite/all-elite paradigm was what I had in mind when I mentioned wanting to rework how elite tags work a few weeks back.

As to reworking mage gen so paths will influence gear... I could very much get behind that. Over the spring I'd been picking away at a very large addition to the standard mage gearset, and it would be quite appealing to make all that new junk correspond to paths. When I say very large, I mean I've already prepared graphics (but not data) for... 28 more miscellaneous mainhands, 12 offhands, and 26 scepters; I still need to clean up the staves, so I don't know how many I'll end up with, but I identified ~150 units as having a staff that I thought was potentially distinct and interesting so that probably means I'll end up with 20-40 of those. I've also got ~80 new pieces of cleaned up mage headgear (recolor variants will easily get that to 100+) that I just need to prepare stats on, and had identified around 200 units with potentially interesting robes (which will probably get winnowed down to 20-30 new robes at the most). Which is to say, if I had the capability to have mage gear reflect their paths, I think we could still manage to routinely end up with a very nice, varied set of mages even within a given path, at least for the size-2 units using the standard gearsets.

PR #25, opened 2015-06-09
Elite bugfix and tuning

elmokki:


elmokki:
These sacredgenerator elites may become rarer if the rostergenerator starts to pour out elite units properly at some point.

PR #27, opened 2015-06-09
Pulling recent theme changes to master before starting more major work on unit generation

elmokki:


Ealb:
A question on how lockslot would work. Basically what I'd be looking to do would be to be able to, say, #lockslot basesprite humanbred_base. However, with that if I wanted a humanbred mage to be able to use different hands in the cases of different gear, would it be possible? I.e., if they have a staff they'd use the basic hands_humanbred_default w/o problems. If they were holding a flame or bowl in their hands, though, they'd want the #nogen hands open_palm that this item #needs. Is this currently possible? I.e., is there a way that I can specify that the unit needs a slot foo_slot with the name foo_item, and it'll e.g. only consider the items of that slot with the name foo_item and theme #humanbred? Basically, I'm trying to see if there's a way that I could have subraces use varied hands/legs/etc. that are specified as #needed by armor.txt/offhand.txt/etc. without giving them custom humanbred_armor.txt/humanbred_offhand.txt/etc. which explicitly states #needs legs plate_humanbred instead of #needs legs plate. If e.g. the humanbred baseprite could have #needstheme legs humanbred #needstheme hands humanbred, and then whenever legs or hands were selected only those with the right theme would be considered (whether it's a random selection out of a list of six legs, or it's explicitly looking for "plate", and there's both "#name plate #theme abysian" and "#name plate #theme humanbred") that would be as close to ideal as I can see. Is something functionally equivalent to this currently possible?

Ealb:
"Some subrace mixing thing might happen later if I figure out a clever way to do it: Probably via some sort of subrace definition thing. After all, it would be nice if we could have different tribe Caelians in the same nation occasionally since they are visually very different. Though technically this is doable with just separating Caelians to separate races as well, so I don't know whether it warrants the extra complication of definitions (codewise it should be relatively simple)"

This is actually kinda what I'm trying to do with the above muddled comment. If we have a basesprite_humanbred that is able to specify that it needs "#theme humanbred" hands and legs, or raptor wings, it should be possible to have a nation pick and choose between various basesprites and the corresponding correct subrace-specific bits of anatomy would just sort themselves out. I think it'd be worth the complications to definitions if for no other reason than it would cut down on the number of variant definition files running around; the fewer redundant definitions we have, the easier it is to comprehensively fix errors or update items following stat/system changes.

Ideally, I'd only had separate race variants for instances where it makes sense for different races to ally with one subrace but not another (and the subrace<->secondary half of that we can already do with themes), as I'm pretty sure this is the only thing we can't actually do via themes. I.e., Machakans ally with lizards more than other humans, so being able to specify that lizards have a higher chance of Machakan secondary races over humans is desirable, so we wouldn't want Machakan to be made a subrace of human (at that point, Machakans could have secondary lizards at a higher rate, but lizards couldn't have secondary Machakans more than other humans). OTOH, Machakans ()and everything else) ally with both predator and herbivore lizards at the same rate, so there's no real need to maintain predators as a distinct race, and I could probably safely fold sobek.txt back into lizard.txt and instead have 3-5 lizard tribal themes combining different mixes of poses for herbivores, slave predators, free predators, and sobeks.

elmokki:
Okay, I have some issues in understanding exactly what you are asking. So, is it basically summarizable by a case below?

We have Muuch. They have all sorts of different skin colors. 
1. You have a theme with #lockslot basesprite muuch_yellow for example
2. Muuch weapons have #needs hands 2h_hands for example, but this is the default blueish green color hands
3. You can't really make yellow skinned Muuch without making weapon variants that have #needs hands 2h_hands_yellow.

If this is it, yeah, not possible now. I did not think of a case like above since none of the races I've made have hands in #needs but instead just generated in poses. When they are in poses you could just #lockslot hands yellowhands and have hands named yellowhands in every different hand definitions or do the same and use chanceincs instead of #lockslot. Making it work for the above case is bound to be far more complex than the current implementation. 

Now, as for subraces:

"This is actually kinda what I'm trying to do with the above muddled comment. If we have a basesprite_humanbred that is able to specify that it needs "#theme humanbred" hands and legs, or raptor wings, it should be possible to have a nation pick and choose between various basesprites and the corresponding correct subrace-specific bits of anatomy would just sort themselves out."

This is what the poses are if I don't understand you wrong. I mean, the shamblers in Muuch and Atlantians limit the anatomy by being different poses. From what I understand your idea would be to scrap the current pose system so that technically you can merge shamblers and smaller units to same pose (even if it is not the goal here). 

I understand that you want to merge Sobeks, Lizards and slaver Lizards. For slaver lizards and lizards this is almost definitely already possible with themes since the differences with those races are minimal. For Sobeks handling them as less than a race but more than a set of poses in lizards does make sense since while they should occasionally be on their own leading a pile of some completely other race, they generally are something that should potentially be part of lizards even if they have human slaves or something. Most of this stuff is possible as poses, but too random. The way I see this it doesn't contain any of that needs "#theme humanbred" stuff though, probably because I just don't get your idea.

Why did you feel that sobeks and humanbreds need to be separate races by the way? They are completely legit choices, but seeing how you want to merge them and I'm not 100% clear about your idea on how, I'm interested.

I guess my main problem is that I really don't get the implementation you are suggesting and how it'd make things simpler. You really need to give me simple examples here. 

Most redundancies come, I believe, from item definitions since essentially many races have the same weapons and armor but with different sprites. However, since the probabilities and possibly themes to an extent vary, I'm very vary of trying to concentrate them without a very solid idea about a great implementation. Races and poses have, I think, fairly little redundancies.

Ealb:
The main reason I went with a different race for sobek was because at the time I couldn't conceive of how else to do it. I'd say at the time, I couldn't have done it any other way w/o having extremely low cohesion. With humanbred, it was moreso because I didn't really think to just make separate poses for them, I suppose, and that would have been the only way that "#needs legs foo" would have worked for the armors - although in this particular case I could have just done what I've done with more recent races, and make the legs slot only give the pants/armor instead of pants/armor-and-legs and use a basesprite that gives the full body. That doesn't work for hands, obviously.

With sobek, all the poses are entirely distinct and don't have meaningful overlap. They have some identical gear, but their armor isn't sharable because of extreme morphological differences. Humanbred, OTOH, are essentially the same as Abysians except for the basesprite. They use the same gear, they have the same roles (well, except for ranged) which use the same gear (hmm, well, so long as I tagged all the leather stuff as humanbred - I hadn't thought of that part). With humanbred, I probably could seamlessly merge them with the tools we have now and just make variant poses that are almost identical except for having different bases.txt/legs.txt/hands.txt. It'd be redundant, but it wouldn't be extremely redundant. With muuch it would be more problematic, because they have more variants (4x colors) that would repeat poses identically except for their base/hands... but even that wouldn't be an enormous problem. Atlantians might be a bit worse just given how many pose variants there are, but come to think of it themes reduce the need for rigidly defined "ice gear" and "meteorite iron gear" poses, so having e.g. deep one vs. reef variants of every pose might not be so bad if the poses get consolidated, since it'd just be "reef infantry" and not "reef coral/ice/iron/bone infantry", and gear themes could take care of the rest.

The more I think about this, the more I think I can make do with what we have now. #needstheme would allow me to have a bit more control (especially in terms of encouraging coherent specialty gearsets; e.g., if you have a fancy silver breastplate you don't end up with a golden helmet), and to reduce redundancy a bit, but for subraces I think I should be able to do fine by specifying poses. The only thing it would miss would be the ability to just let basesprite selection hold all the variance for subraces, which would make it much easier for low-probability random subraces to pop up in homogenous populations... but even then I could just include out-of-theme subrace poses at a very low probability if I wanted to achieve that effect.

I'm going to say disregard my above request for more functionality along these lines for now. If I find that I can't really do what I'm wanting to with the tools you've given me right now, I'll come back and whine for more, but the more I think about it, the more I'm convinced that just making more poses will achieve the same effect that I'm looking for w/o requiring code changes. The only thing I can clearly see this preventing me from implementing is a ridiculous number of ichtyid pigments; if I could just specify #needstheme hands foo1 in basesprite_foo1 instead of making infantry_foo1.txt plus mages_foo1.txt, ect., I'd cheerfully use all of the 9 or 10 scale pigments I made, but I don't want to create that much clutter for so little in return.

Ealb:
To expound on/clarify the above, what I'm thinking now is that I'll use #lockslot basepsrite for icthyid subposes. With Abysians and Humanbreds, I'll probably make the race file load few or no poses, and then specify a #freetheme tribal (or maybe demographic; it doesn't really matter) and then use those themes to load different combinations of humanbred/Abysian troops/mages (e.g., primal: all Abysians, like the current abysian.txt; degenerate: all humanbred, like the current humanbred.txt; somewhereinbetween1, somewhereinbetween2: somewhere in between). And techlevel primitive would reduce humanbred pose chances via themes, just as advanced would reduce Abysian chances.

Lizards would likewise load few or no poses by default and then have a mandatory tribal/demographic theme mixing herbivore, slave predator, free predator, and sobek poses in varying ratios.

Yeah. This should work well.

elmokki:
I think we have two really easy solutions to the basic issue here.

First is to add #needstype slot type that you can use to have #needstype hands 2h_hands to choose a random hand with #type 2h_hands. Then you can use #themeincs or #chanceincs to make only correct color hands happen. To top that it'd be pretty easy to add #chanceinc item slot name so you can just add the yellow hands #chanceinc item basesprite basesprite_muuch_yellow. The second is to just add #needs_if_slot basesprite basesprite_muuch_yellow 2h_hands_yellow. However, I think the first option is better. 

Both are about the same complexity to implement both in code and content. However, the first one only requires you to add one line that you will not need to edit to add more variants to each item while the second needs a line for each different basesprite. On top of that the first one is so very general that it's probably possible to figure out more innovative uses for it. Of course the easiest for content making would be to have that kind of #restrict_slot_to_theme slot theme, but it's far less messy to have the potentially available items for a slot be affected by the potential items instead of all the other items. 

So, yeah, I guess the first option is the best?

Overall I completely see the issue with #lockslot: It makes all the units in the race use the same pigment or wing color or whatever. Caelians should definitely have nations with multiple tribes occasionally and aren't there distinctly statted Muuch of different color as well in vanilla? I think subraces are something I should really consider before rewriting  troops and mages since the logic allotting subrace to units would probably be fairly major. There is also the question of how to exactly handle them, since some subraces would just consist of different poses (Sobek, Shamblers) while some would consist of shared poses with different #lockslot (Caelians).

Also, technically there are no visible issues in adding races or poses or something (or both?) a list of colors and slots to recolor with specific colors. If the difference in the basesprites and hands is attainable by recoloring items just like recolormasks work now, that's definitely handleable without sprite editing. The biggest question is about where to add the recoloring.

One technical idea that also sprung into my mind is that technically it would probably be possible to have for example a caelian pose without basesprite and basechance 0. Then other poses with command #copypose base_caelian and the basesprite and other stuff. However, this may open a can of worms with how some of the commands are handled. Not that the same can of worms wouldn't already be open with themes and race commands though.

elmokki:
Also, using #freetheme to load poses sounds extremely ingenious and I don't see why it wouldn't work. However, if it doesn't, you can always just #theme poses and use #themeincs. They should work with all pose selection code now and if they don't, it's a question of finding where they don't work and changing like one line for each case.

elmokki:
...and  one more comment. Atlantians were separated due to how the generation code has worked for a long time: It tends to select a pose and make a pile of units of that pose and in addition you want to generally have ice weapons with ice armor and so on. So basically combining the poses has the potential of having a huge mismatch in your roster. 

I think that for the rewrite I'll utilize #type instead since it's a bit dumb that it tries to make 3 polearm and/or 1h + shield humans and then separately add twohander humans. It's better if the basic melee humans that have all the same armor are considered the same even if the poses are different.

Now that I think of it, it'd also be possible - in a rewrite of pretty much all unit generation - to add code for pose variants which are actually pretty much what you originally requested. This'd work like #posevariants variants_atlantean and then variants_atlantean has ice, coral, reef and meteor which all just remove items not tagged with their own tag. However, as I said, this will need going through the code and writing a lot of stuff again. All the stuff that does use unit in determining chanceincs would easily support this, but I know everything does not. 

This doesn't replace the idea of subraces due to subraces allowing far more control in how the subraces are distributed in generation, but it does allow merging pose files to one like you originally wished. I suppose subraces can utilize posevariants to force the use of certain stuff though.

Ealb:
For Caelian nations that aren't homogenous, I was thinking of doing as I described with lizards and just running through several permutations of pose combinations. However, that's narrow and kludgy, and it would work better if I could just modify wing color chances based on theme and use that to come up with a mostly-foo1, sometimes-foo2 mix of Caelians. They don't have issues with hands, so that'd be easy. For mixed races of muuch. or mixed races of traditional Atlantians, I was planning on using different themes with fixed combinations of poses (to include some "all poses, but foo1 is weighted higher" themes, so a chak muuch theme would mostly give you chak muuch, but you might have some brown or green ones thrown in as well) - honestly I'd probably have just gone with a paradigm of [foo1x10, foo2/3/4x0.5], [foo1x5, foo2/3/4x1], [foo1/2/3/4x1] chance breakdown via theme tags on the poses. I still think that could work, and likewise for Deep Ones, but it might be cleaner in this case if I could just go with some sort of #needstype and then just rely on picking subrace basesprites instead of subrace poses to give us a mix. Either way should work, though.

I just remembered the other reason I really wanted #needstype, and I'm not entirely sure it's necessary. Like I said, I have a pile of mainhand/offhand objects for mages, and I'd like to update the main default files with them. However, some of them will definitely look better or possibly even only work if the hand is open instead of closed (again, holding a flame/bowl/ball). I *think* that I can weasel around this with a simple #needs hands_openpalm or whatever, and then going to the hand file of every race that uses generic mage items and adding a #nogen item named hands_openpalm selection of the appropriate size/shape/color. That should work. Otherwise, though, being able to type certain hands as #openpalm would be useful. If I wanted to get really tricky and not make separate lhand rhand slots, being able to specify #needstype openpalm_l in the offhand and #needstype openpalm_r in the mainhand, and then having the program pick out hands_both_open over hands_left_open, hands_right_open, and hands_both_closed based on them having both, one, or none of the tags would be very nice. But. Again, I don't think I really *need* this level of functionality, and can make do by either imposing more uniformity among items, or by making separate slots in cases where something like this might occur.

I don't know if the above paragraph was entirely clear, but it pretty much comes down to "I don't think I need it, but it'd be nice and encourage me to make more (needlessly?!?) complex definitions". If you think it will be reasonable to add it during the re-write, that certainly works. If not, I think I can work around it pretty well. I would say that I'm wary of just doing re-coloring for subraces whose visual distinction is just color. It's the most appealing and least redundant possible solution, but some of the canonical subraces of this sort have odd combinations of different colors on different body parts, and it'd be safer/cleaner to muck about with sprites.

elmokki:
Just #needstype as a command that is essentially exactly as #needs but is instead a chanceinc weighted random selection from #type restricted set doesn't need a rewrite. I can whip it up once I've eaten.

PR #34, opened 2015-06-12
Elmokki

elmokki:


Ealb:
I've been sore-tempted to pull a few more icons and such out of Dom's flags at some point; I may give it an hour of attention over the weekend.

I had some thoughts about themes interacting with filters, and I'll probably test those tonight. I'm not going to mess with the probabilities on the slightly-too-rare themes myself, at least not right now, but the tentative plan is to do a sweep through all the filters and identify "animal-ish" ones (and maybe add a few more mimicking the ones we already have) and drop theme tags on them. I have two competing notions of how precisely to do that, both of which are relatively straightforward, but I'll see which works out more smoothly. The idea will be to allow tribal themes to e.g. increase the likelihood of e.g. having a battlesummon wolf, or turning into a (were)wolf, or being a wolfherd, etc., while possibly reducing (but not eliminating) the chance of e.g. turning into a toad - or vice versa if toads are your nation's "thing".

Ealb:
Oh, on the subject of filters, it probably wouldn't be the worst idea to look at maybe expanding the list of second-shape inheritance for non-mounts to include most or all of the bad traits. Right now if you both shapeshift and cause unrest/act as a heretic/etc. you can just change shape and cease to be a bad neighbor at the cost of a bit of research potency.

elmokki:
Should you do a filter sweep, do it the way I suspect you were going to do it anyway: Just tag filters with "wolf", "toad" etc. and add #themeinc theme wolf *2 to wolf tribe and so on. Now, this way you can't reduce the probabilities of the non-wolf animals, but if you also tag "animal" I guess I could add some complexly named #themeinc hastag_but_nottag pretty trivially. 

This is because if you start adding #chanceinc hastheme "wolftribe" *2 stuff to filters themselves they're gonna become very bloated over time as more stuff like this is added, while tags like "wolf" and "animal" are probably reusable. Also, reusable not only for filters and themes, but also for naming for example. I think you can already use a chanceinc to make a name more common based on tags, so we'd have a more general way of making for example the name wolfkin happen with wolf-related stuff. Currently I think it checks for shapeshift to werewolf so it's very rigid. 

The shape inheritance is in a file so adding a few bad commands there is trivial.

Ealb:
I went the slightly-less-elegant route of adding the theme "wolftotem" and "totemanimal", which seemed generic and reusable, and I could have each tribe increase their own totem while decreasing the general totemanimal. It works, mostly, although you still end up with as many or more non-totem mounts and shapeshifts as totem ones, which is where my "maybe it's not enough" comment comes from. I like some variability, and you can usually see some influence, but I may tweak it slightly higher, even if perhaps only indirectly by adding some more mounts (e.g., I'm seriously considering working out stats for "normal, not-so-great bears that'd be maybe size 2 or 3, which would be a better shapeshift for smaller mages/elites as well as a better mount for smaller races), as theme "wolf"/"wolftotem" has two varieties and actually is fairly influential mount-wise.

Incidentally, it's Very Bad to use any theme that starts with/is the word "animal". The "animal foo" theme for mount descriptions really, REALLY doesn't like it...

elmokki:
Yeah. #animal horse #theme "animal horse" and #tag "animal horse" result in the same thing. Ideally we'd have #theme separate from the other two and #theme would be for thematic tags and the other way for more technical things like signifying that the mount is a horse.

I might go over it and change any #theme that isn't obviously a thematic keyword to #tag and then separate the two once again.

elmokki:
I mean, after all, it's probably mostly the mounts where #theme is used instead of #tag in a way that will break things if the two are separated.

PR #50, opened 2015-06-15
#tag "elite" -> #theme "elite"

Ealb:
- Updating elite tags to themes
- Verifying there weren't sacred tags


elmokki:
Great! I set the issue mainly to remind myself, but this is definitely okay

PR #53, opened 2015-06-15
Actual roster update

elmokki:
The old one got stuck with resolving conflicts.


Ealb:
Oh yay! I did miss preview generator.

I'm doing hoburg themes right now (it was time), but I'll look the roster stuff sometime today.

elmokki:
The preview generator is pretty much Dom3 code I dug from 2012 with few tweaks to get it show mounted units, line up the new larger flags and to have it toggleable in the GUI. 

The #themeincs in filters are something I really think has potential for funky things. I plan to add those template kind of filters available for sacreds and mages with one or more being given always, so you could make thematic sacred templates that happen more often with certain themes - like a bear tribe often generating a sacred with very high themeincs for suitable gear and filters.

elmokki:
Also there are some very definite quirks. I'm looking at three identically geared predator lizards. I might have to do some very heavy rewrites to the rewrite.

Ealb:
The identical units would happen periodically under the old system (or even more often, identical except for something superficial like shirt or hair), but it wasn't at all common.

elmokki:
Now it is pretty common. So yeah, something is very wrong.

PR #55, opened 2015-06-15
Avvite name  fix

elmokki:
They were using human names. Also, I think regular human names might be
better off made germanicish (but distinct from norse and hoburg names)
rather than random gibberish.

### DIFF
--- a/data/races/avvim-slaver.txt
+++ b/data/races/avvim-slaver.txt
@@ -46,9 +46,9 @@
 #pose avviteranged
 #pose avvitemage
 
-#longsyllables /data/names/nations/human/longsyllables.txt
-#shortsyllables /data/names/nations/human/shortsyllables.txt
-#suffixes /data/names/nations/human/suffixes.txt
+#longsyllables /data/names/nations/avvite/longsyllables.txt
+#shortsyllables /data/names/nations/avvite/shortsyllables.txt
+#suffixes /data/names/nations/avvite/suffixes.txt
 
 -- Always have shirt
 #generateitem 1 shirt
--- a/data/races/avvim.txt
+++ b/data/races/avvim.txt
@@ -34,9 +34,9 @@
 #pose avviteranged
 #pose avvitemage
 
-#longsyllables /data/names/nations/human/longsyllables.txt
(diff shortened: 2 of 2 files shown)


Ealb:
Ha, given how much of an ordeal I hold making placenames to be, it's funny that I went to the trouble of making Avvite placenames and then forgot to turn them on.

I agree in re: generic humans. I can probably muddle through that at some point if you don't want to, but if you speak a Germanic language outside of English you could probably do it with more finesse (for the record, I'm a native anglophone and rusty-but-fluent francophone, although I additionally have some formal education in linguistics which certainly doesn't *hurt* for a task like this). Well, to some degree - I certainly don't speak Greek but the "Amazon" names seem pleasing Greekish.

Of curiosity, are the Agarthans gibberish or are they aiming for some specific linguistic origin? They and a couple of the other really small datasets could probably use expanded some day, although I'm not about to suggest it's a high priority.

elmokki:
I think the Agarthan names have some very remotely North Indian-Nepalese thing going on. I did not write them though, kianduatha did. I think they are in general pretty fitting for Agarthans.

I know a few words of German and I speak barely passable Swedish, but I don't think those have really helped that much for hoburg or nordic names. Searching for place name lists and writing parts based on those is how I've done it.

Ealb:
Hmm, I'm more familiar with South Indian languages than North, but just a region should be enough for place-name based muddling, which has been my main method as well (although I did get fancy for the Marignoni humans and specifically dug up French exonyms of Spanish and Italian placenames to make it more indistinctly Romance). 

Another option for generic humans might be to go with Anglo/Scots Germanic names; i.e., equate them with Man moreso than Ulm. Although getting non-hoburgy Germanic placenames probably wouldn't hurt anyway if I drag out barbarians at some point here. Though hmm, I suppose that the opposite argument could be made and barbarians could take Gaelic placenames. Hmm.

elmokki:
It would be possible to just write name sets and pick a random name set from them, like, make generic humans pick from a pile of possible names. It's a pretty niche feature to add, but also fairly easy one.

Equating the basic humans with Man does make sense due to the basesprites being directly from Man.

PR #108, opened 2015-07-07
1st pass through naming data; WIP hoburg stuff

Ealb:


elmokki:
You need to use #chanceinc slotname instead of #chanceinc slot if you use item names

PR #129, opened 2015-07-11
#dw_maxlength and debug message removal

elmokki:
#dw_maxlength sets the maximum length of weapon to be eligible for an
offhand weapon. Default is 2.


Ealb:
I think this broke the prior commit, as it appears to be deleting the code that checks for "#ignore_dw_restrictions" - it looks like that got deleted instead of the debug message right underneath it. It's also trying to pass unit.getSlot("weapon") into isDualWieldEligible(Unit u). I'll commit fixes for both of those issues.

PR #132, opened 2015-07-12
Austral hoburgs base graphics and data

Ealb:
- Done with visually-distinct-but-statistically-indistinct austral (quasi-Machakan) hoburgs
- I'm trying to think if there would be a downside to setting basesprites to include e.g. "#needstype weapon austral" "#needstype armor austral" "#needstype helmet austral" "#needstype offhand austral" and then just using the same pose for all non-oriental hoburgs (for oriental I wanted to screw with DW chances, so I'm okay giving them separate poses - plus this way there's a 1/1000 chance of them randomly showing up in other regional themes, and vice-versa) while simultaneously going through all the hoburg item files and adding "#type austral" to everything that is "compatible" with the austral theme. It *seems* like it would be a relatively clean way to end up with distinct regional subcultures while minimizing redundant data definintion...


elmokki:
The main downside of #needstype like that is that it would require rewriting large parts of code that handles equipping. Parts of that code is of course pretty easy since it uses methods like getSuitableItem() so you could just edit that, but a lot of it doesn't.

There's an alternative that should do close enough the same for less code: We already have #themeinc thisitemtheme and some other such things. Since items actually do store what slot they're for, you could write a themeinc that checks for slot and type. However, it's a bit clunky if it needs to affect one pose only. You'd have to give a freetheme in pose definitions and affect the probabilities in your national level theme.

elmokki:
I'm honestly a bit doubtful of whether it is better to not just make the poses different when the gear is different.

Ealb:
That's a reasonable assessment; I think it's fair to say that my desire to do what I described above is a direct manifestation of my worst impulses towards "efficiency" and minimal data definition. It occurs to me that the only reason I'm having problems just using themes as you suggest is because I was still approaching basechances as though there was a default separate from the regional themes, but since they're mandatory there really isn't. If I just go back through and treat boreal as one choice among others (even if it'll still have a higher chance of occurring than the others) rather than "the default", just using themeincs should do the trick quite nicely.

elmokki:
Even if there is a default you can set the non-defaults have basechance 0.01 or something and then just multiply them by hell of a lot. 

In any case where I wouldn't want to leave some chance of getting cross-cultural gear I'd just define them as separate poses. I strongly prefer using separate poses for most cultural stuff because that guarantees that the whole gear is of certain culture. It depends on the graphical styles, but I wouldn't want a bronze cuirass with an iron cap or a bronze sword with iron gear. However, oriental armor with a western weapon or vice versa probably fits pretty nicely. This is just my opinion though. Stuff can be strange if it's rare enough.

Ealb:
I'm leaving the oriental as separate poses, and scaling back the multipliers from *1000 to *100. There should be some random bleed-over, but not much.

What I had been doing was treating boreal as default and giving it the values I actually wanted, while non-boreal would get 1/1000 of that. Then boreal had no theminc, while the others had themeinc boreal *0.001 and foo *1000. That worked well enough, as even if I wanted some overlap the decrease/increase that happened when something was both boreal and foo kept it balanced, but it got a little weird with how I handled clockworks; I had oriental clockworks showing up in all cultures because there was an absolute increase in chance for the mechanical theme, and my final basechances were low enough that +0.25 or +1 made a difference. If boreal is getting *100 like everything else so the final basechances are ~100 instead of ~1, this problem should disappear.

PR #137, opened 2015-07-13
Minotaur tweaks

Ealb:
40g -> 30g;
+2 att/+1 def;
civilized split into hoplites (formationfighter, +1 mor, spear/trident/longspear/pike) and other
soldiers (trample, +1 def, any weapon w/ DW possible)

### DIFF
--- a/data/items/gaian/minotaur/2h_bases.txt
+++ b/data/items/gaian/minotaur/2h_bases.txt
@@ -1,5 +1,21 @@
 #newitem
-#name "basesprite"
+#name "savage"
 #gameid -1
 #sprite /graphics/gaian/minotaur/minotaur_basesprite_2h.png
+#command "#berserk +4"
+#theme "savage"
+#theme "primitive"
+#enditem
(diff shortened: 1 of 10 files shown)


elmokki:
I think it might be a good idea to have other onehanders without trample as well if they are carried with a shield (and #formationfighter itself isn't necessary), but this is probably great for balance already so do what you wish.

This does remind me that adding a troop template #formationfighter and tagging some spears etc suitably would be a great idea.

Ealb:
I actually had been thinking that letting shieldbearers trample might help up trampler survivability a bit, even though the added encumbrance will of course hurt them. Hmm. Though you're probably right that it wouldn't be the worst thing to allow some more. I might try re-tweaking this, or I might not - I do suppose it would be good for me to go ahead and finish my 0.6.0 milestones instead of continuing to distract myself with random changes...

elmokki:
It might be a good idea to just release as is and see feedback.

PR #144, opened 2015-07-14
Non-stealthy hoburg wings

Ealb:
Moderately low probability bat wings for the more civilized/non-fae sacred hoburgs (~1/4 the frequency of wings for human sacreds, when applicable)


elmokki:
This actually reminds me: Gotta copy those icarid wings and recolor them and add some very low chance of icarid units for various races.

PR #150, opened 2015-07-16
Hoburg cultural theme distribution/homogeneity changes

Ealb:
- Messed about with proportions of cultures in each techlevel
- Increased cultural theme homogeneity by a factor of 10 (most stuff) to
100 (bases in particular)

(There seemed to be a distressingly greater proportion of random "exotic" cultural units thrown into any given hoburg culture than its ~1/100 chance lead one to expect, particularly but not exclusively clockworks and scouts)


elmokki:
It would've been pretty easy to make the program write all the chanceinced probabilities and see if they're messed up.

elmokki:
I generated 10 nations with whatever I had pulled, which is before your restructuring, and I did see one dark skinned hoburg amongst asian hoburgs, but other than that it seemed just fine.

You've really made excellent work with these. You could instantly see when a nation was advanced and when it was oriental etc.

Ealb:
Where it got weird looking was that it would happen in clumps a lot of the time - which I'm pretty sure was an artifact of the way units are picked and varied from a single pose. So you'd e.g. see a boreal nation with 3 hoburgbred clockwork troops - or a couple of bronze (i.e. occidental/central) clockworks in an oriental nation. I think the oddest I saw was a boreal nation with a pair of austral hoburg units in oriental clockwork armor. I'm not sure what was up with scouts, but when I'd generate 40 units I'd usually see 3-4 either with a different base than the rest or not matching up with the rest of the nation in terms of oriental/occidental. Running a set of 100 would consistently see ~5-10 "mischosen" scouts, where statistically it should have been closer to 2-3 on average. This latest rebalancing/overtuning/what-have-you is spitting out hoburg nations that approach the homogeneity of human subtypes without requiring completely separate race and item files (and with mostly overlapping pose files), so I'm happy.

I'm still looking to do a pass through and make a handful of graphics for bronze gear and such, which should further distinguish tech levels and culture. I am pretty pleased with how the mages for various cultures are actually somewhat distinctive, even if it's subtle (and might only be apparent to me because I know what to look for...)

Ealb:
(It might still be worth my time to look into the chanceincs when I'm done with my current block of content, though, as I merely sidestepped any potential underlying problem with my "solution" here.)

elmokki:
Yeah, I committed a fix to this a moment ago for weapons at least. The program loves to use gear given to a differently armored unit for a new unit because why not - it makes nation cohesive. I made a check where if the probability mass of the old stuff is less than 5% of the new stuff a new item is forced. 

The number needs to be somewhat low so it won't trigger too easily for nations with legitimately large amount of gear, but I ran some checks and this seems to be okay. It's worth a note that even if this triggers it can still generate a new copy of a very rare used item - it's just gonna have to compete with the less rare stuff.

elmokki:
...and same fix for other stuff as well.

As for scouts, generating hoburg scouts - before your restructuring - gives this kind of chanceincs for pose selection:
{[infantry, scout] (boreal infantry)=10.0, [infantry, scout] (oriental infantry)=0.1}
Basesprites, similarly:
.{hoburg basesprite=10.0, austral hoburg basesprite=0.1}

Ealb:
Huh. So it really was just improbable odds over several test cases. It was only off by a factor of ~3-4, so it's not completely incredible, just surprising. I still tend to think the nations are less likely to be odd looking with the extra decimal place from the restructuring, though. ~1/100 (or in the case of scouts, closer to ~1/50) seems like it's still low enough to see multiple (different) instances crop "occasionally" in one nation, and that should probably happen "almost never".

elmokki:
Do note that your restructuring may have affected those

Ealb:
Oh. Oh my. If those are the post-restructuring numbers, I'm suddenly surprised I was getting as few random exotic units as I was, as the restructuring should have increased the proportions by ~10x. I suppose ~10 "wrong" scouts out of 100 sounds fairly close to reasonable for ~10x more than what those numbers are currently showing.

elmokki:
Pre-restructuring

PR #208, opened 2015-07-26
Heavy clockwork rebalancing

Ealb:
+10 rcost
#sacredratingmulti *0.75
#sacredcostmulti *1.25
Fear *1.2 gcost -> *1.25 gcost
Minimum 0.9 chance of caponly

### DIFF
--- a/data/items/hoburg/clockwork/armor_clockwork_heavy.txt
+++ b/data/items/hoburg/clockwork/armor_clockwork_heavy.txt
@@ -9,7 +9,7 @@
 #theme "advanced"
 
 #define "#gcost +40"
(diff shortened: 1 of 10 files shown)


elmokki:
I think some restrictions on what sacred stuff they can get should actually happen as well. I can do it eventually. Shapeshifts in general, except maybe the elemental ones, don't make a lot of sense and some things like glamour and maybe ethereal are a bit too much.

However, I kinda think that first I'll divide the sacred filters to a pile of separate files so we won't have so much redundancy. It's really stupid how now many filters are in three different files. A slight code change would allow having any amount of defaults if no specific filters are specified anyway.

PR #209, opened 2015-07-28
Gave several shapechange filters increased #maxage 

Ealb:
Devils, vine/metalmen, deepone to full age of second form; werefoo *2

### DIFF
--- a/data/filters/default_elitefilters.txt
+++ b/data/filters/default_elitefilters.txt
@@ -192,6 +192,7 @@
 #chanceinc "shape werewolf 20"
 #command "#secondshape werewolf"
 #command "#gcost +10"
+#command "#maxage *2"
 #chanceinc "pose mounted *0"
 #chanceinc "pose chariot *0"
 #power 3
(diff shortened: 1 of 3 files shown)


elmokki:
There is code that sets maxage to the same for both shapes based on primary shape. It will probably override your changes.

This was because on some Dom4 version the game crashed if the shapes had different maxages. Not sure if it was changed.

elmokki:
The code should be in ShapeChangeUnit

elmokki:
#maxage *2 will also only work when #maxage has been explicitly defined in NationGen, so it only works if the #maxage code thing happens before resolving those commands.

Regardless, do test.

Ealb:
I went ahead and set these in the filter, not in the secondshape, so it overrides to the higher value.

I'll test the werefoos though - that does seem like a spot that's ripe for problems.

elmokki:
If it's in the filter I suspect that it'll only work when the unit in question has #maxage specified when it's *2 and not static

Ealb:
Aye, I'm suspicious that that's the case. It could be kludged by defining maxage for everything but that's an awful solution for a small amount of gain.

elmokki:
The program already sets some default commands for races afaik. If it sets #maxage 50 which iirc is human default there that would solve it

Ealb:
Aye, that's a clean, easy 1-line fix in Race.java

Ealb:
Ha, and it's definitely needed because without it Dom4 interprets maxage as 0 so you end up with maximally oldage units upon recruitment.

PR #213, opened 2015-08-03
More hoburg stuff

Ealb:
Preliminary vine/pod hoburg stuff (currently disabled)
Added some tags to some filters to allow for "ivy" associations
Ermor/Abysian-style armor for hoburgbred
"Imperial" hoburgs as a distinct cultural theme


Ealb:
I probably need to go through and define an Imperial placename file. Though hmm, I suppose imperial hoburgs can use "Romance" prefixes with hoburgy suffixes as a stopgap

PR #253, opened 2015-08-12
strongmagicpatterns for low-tech hoburgs only

Ealb:


elmokki:
This wasn't enough. 

The magic patterns have chanceincs such as `#chanceinc racetag strongmagicpatterns 4` which target racial tags only. Currently magegenerator only checks race tags as well. Fixing this.

Ealb:
Bah, ofc, why else would you have mentioned chanceincs.

PR #261, opened 2015-08-28
Ichtyid mages res penalty/mr bonus

Ealb:
+2 mr/-2 res (in line with talented Pelagian ichtyid mages rather than
the "how did these things learn magic?" indy shamans)


elmokki:
-2 is kind of bad if the nation's main mages get it though

PR #272, opened 2015-09-07
Added colossi elite/sacred troops to austral hoburg theme

Ealb:
* No, I don't mean size 3 colossi
* Yes, I should be ashamed of myself
* No, I'm not

(This installment of "But don't you have *better* things to be working
on?" is brought to you once again by the theory that "Life underneath
the waistline corresponds in many ways to that at eye level.")


elmokki:
This is hilarious.

PR #278, opened 2015-09-15
armor.csv field name error

Ealb:
Not sure how this happened (DB conversion stuff doesn't point to an
immediately obvious explanation, either), but "armorname" had been
renamed "name" in armor.csv, so Advanced Description files, etc. were
all screwed up.


elmokki:
It was name in modinspector files and was left there accidentally I suppose. 

Unlike Dom3/4db the modinspector databases use actual dominions command names for fields. Damn that would've been useful from straight at the start (I have some code to convert these for example, which is such waste)

PR #292, opened 2015-09-19
Caelian armor

Ealb:
First pass at distinct Caelian armor
* Still need "magic" armor for Yazata/Daeva
* Need to add chanceincs to ensure that ice armor doesn't occur w/o
coldres
* Need to test to make sure chances and themes are enough to keep the
various types from being too diverse with the various clans all using
the same poses
* Need to check/add helms and shields
* Add some oriental armor? Too much like tengu, or too much effort?
* Need to add occidental armor, but that can wait
* Should techlevels between primary and secondary be more likely to
co-occur? I.e., add chanceincs to e.g. boost bronze techlevel chance if
primaryrace is bronze?


elmokki:
If you add oriental armor, it's a good place to make Tengu appear more often. Either as a secondary race or even like Hoburgs have cavemen.

Similar tech levels for both races would make sense I suppose

PR #293, opened 2015-09-20
Monsters were not processing themeincs

Ealb:
When selecting monsters, the monsters' themes were not triggering
themeincs in the nation's theme. This has been corrected, although my
"solution" is better described in terms relating to pasta than those
relating to good programming practices.

### DIFF
--- a/data/monsters/default_monsters.txt
+++ b/data/monsters/default_monsters.txt
@@ -58,7 +58,7 @@
 #new
 #name "elephant_caelian_raptor"
 #basechance 0
-#chanceinc primaryrace caelian 0.008
+#chanceinc primaryrace caelian 0.08
 #chanceinc nationcommand idealcold above 0 *0
 #theme "raptor"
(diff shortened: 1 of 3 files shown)


Ealb:
Actually, that's pretty much what this mess is doing, w/o needing to touch subrace. The actual race of Monsters remains Monster; it's just that the call to handleChanceIncs expects to get a list of units that it'll grab race[0] from, and monsterGen uses a variant handleChanceIncs that *doesn't* take a list of units, but instead initializes an empty list which it then passes in. There was some "it was late, I was tired" going on, but the impression I got was that chanceIncHandler would behave badly if it got a list with stub units who had the correct race set... but looking at handleChanceIncs again, I question that. I may be able eliminate the extra parameter by making the same dummy list of units and then just setting the race of the first one to primaryrace. Monsters will still retain their race, and the clunky fixedRace parameter would be eliminated.

PR #320, opened 2015-10-12
determineSpecialLeadership wasn't catching all special types

Ealb:
* magic/demon/undead units who were e.g. getting their status from an item (like basesprite) were not being properly processed because commands[] hadn't had the command added at the point where dSL() was invoked. A call to dSL() was added for all casters at the end of polish(), although that may well not be a good place to add it - also, it looks like non-caster heroes aren't getting special leadership added, so it might be better if this was done for all non-scout leaders instead of just casters.

### DIFF
--- a/src/nationGen/magic/MageGenerator.java
+++ b/src/nationGen/magic/MageGenerator.java
@@ -1311,7 +1311,7 @@ public class MageGenerator extends TroopGenerator {
 		return priests;
 	}
 	
-	private void determineSpecialLeadership(Unit u)
+	public static void determineSpecialLeadership(Unit u)
 	{
 		String basiclevel = u.getLeaderLevel();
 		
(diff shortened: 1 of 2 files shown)


elmokki:
Actually the smooth solution might've been to make the loop in dSL() run u.getCommands() instead of u.commands. I wonder why it doesn't do that.

I think the ideal would be to make dSL() a static method and make it run for certain things, but then again I suppose making it happen for everyone might be just fine.

Ealb:
That's actually a good point; since it's just giving base leadership, scouts/etc. will still get #no, and any leader who already has better-than-base leadership will keep it, there's no real downside to eliminating the mage check and having it run for everyone aside for a few CPU cycles. I'll change it to that as soon as I track down another bug that came up - some nations aren't getting summaries generated.

elmokki:
chances are it's an error of writing mod file

PR #329, opened 2015-10-17
Metic legs

Ealb:


Ealb:
This is very possibly a large step towards Hellenic/Arco nations, as I always kinda just looked at the metics as "Greeky" men

PR #336, opened 2015-10-28
Naming update, a few fixes, winged lions

Ealb:
* A whole pile of name updates for bronze and custom weapons, as well as some mounts that were missing
* Set standards to add #cannot_be_pd tags and #aisinglerec commands (is there a way to make standard bearers never get selected to be made into commanders? If it's painfully obvious, apologies; it's late, I'm tired, etc. )
* Added winged lion mounts to austral humans and monkey nations - basically flying but more expensive, fragile, and cowardly lions, graphically based on a scaled-down version of the winged lion pretender 
* Some graphics tweaks (e.g., the trim on boots for standard human stuff is a bit duller so it doesn't contrast as much)
* Some data/graphics fixes that cropped up when mass-generating test nations


elmokki:
No way to limit commander generation out of the box. Should be a fairly simple thing to add though.

PR #348, opened 2015-11-02
New .jar for release

Ealb:
* I had a couple of unusual crashes (looked like during hero gen) when I was rapidly going through data changes last night, but I couldn't duplicate them while generating 1500-2000 nations today, so either they were artifacts of screwed up testing numbers, or they're silently lurking in the background.


elmokki:
The best way to debug stuff like that is to just release and wait for someone to report a crash.

PR #405, opened 2016-02-16
Nation themes

elmokki:
So, this involves quite a bit of stuff:

Firstly, nation themes. They work exactly like race themes in functionality really and can use the same commands. They are generated after primary race and primary race themes are generated but before secondary race is chosen. I chanfed #nationthemes to #racethemes to make things more sensible. Nation themes replace nation filters and I replicated them as defaults. Essentially the differences are that nation themes are generated earlier and can affect generation via the normal theme mechanics and any future theme mechanics, so instead of for example giving blood sacrifice conditionally, we now give blood sacrifice and increase probability of blood magic a bit. 

Secondly, I added two chanceincs:
#chanceinc magicpriority path [below] value multi
#chanceinc primaryracecommand command [below/above] value multi
These should be obvious. They're useful for nation/race theme selection chanceincs.

Thirdly, I changed how #magicpriority works. Previously it picked value from pose if it existed and value from race if not and defaulted to 1 if neither existed. Now all #magicpriority commands are multipliers and they affect from primary race racial and national themes as well. Basically if previously you had #magicpriority blood 0.5 on a race and #magicpriority blood 2 on a pose it resulted on 2 on that pose whereas now it results in 1. I didn't seek any values to fix since as far as I know the pose option is unused, but I may well be wrong.

Fourthly, a few new commands to make things in nationthemes work better: 

#magepriestchance <chance>
Chance of priests being primary mages. Defaults to 0.3. There's a chance of secondary mage priests even if this fails.

#highestpriestlevel <level> 
Forces the highest priest level on nation.

#higherpriestlevelchance <probability>
If probability is p, the probability of highest priest level is: p^3 for level 3, p^2 - p^3 for level 2 and 1 - p^2 - p^3 for level 1. Default is 0.5 which results in 0.125 for level 3, 0.375 for level 2 and 0.5 for level 1. 

#priestextracost <cost>
Each level of holy costs this much extra gold.

These all should work in race, primary race themes and nation themes.

Themes also gained
#bothracedefinition which works like #racedefinition and #secondaryracedefinition but affects both races.

Nation themes could be used to replace nationwide filters (stuff like stealth to low enc units or swamp survival for everyone), but I did not do that. I think that might be removed completely and a new category of racial themes should be added to replace it. There are, however, some extra hijinks that might have to be handled for that to work properly, since there is funky stuff like #lowenccommand in the filter-based approach.


Ealb:
The listed percentages for #higherpriestlevelchance are off. Since NG only checks against <chance> twice, it actually breaks down as [1 - chance] for lvl 1, [chance - chance^2] for level 2, and [chance^2] for level 3. I.e., 0.5 means [1 - 0.5] = 0.5 of nations get no level upgrade; of the 0.5 that get an upgrade 0.5 will get a second upgrade so [0.5 - 0.5^2] = 0.25 get a single upgrade; and thus [0.5^2] = 0.25 get two upgrades.

I'll update the documentation.

PR #406, opened 2016-02-16
Work

elmokki:


Ealb:
Was the "lost" longer version #405 ? It's still on the site, just not as obviously so.

I wanna say the only pose I can think of using #magicpriority is yetis right now. I'll check to make sure that's playing nicely with the change.

elmokki:
I merged my branch to master without committing. #406 here is the same stuff.

PR #426, opened 2016-02-28
Crude basic Foulspawn graphic gen

Ealb:
* Enough data structures to generate extremely crude generic fixed-shape
Foulspawn nations
* No weapons, armor, etc.
* No special abilities
* Invariably throws a "can't spawn imagestream"
* Some hanging, some crashing - this also seems to be the case with the
Montag_Men
* This is basically just enough data to make Foulspawn "basesprites"
generate dynamically from basic parts


elmokki:
I'll take a look at this later

Ealb:
In the cold light of day (and having gotten some sleep), some immediate explanations present themselves to me. I'm pretty sure the imagestream issue was because I had stub mounted poses w/o mount basesprites - although I only added the mounted stub poses because RosterGenerator appeared to be looping into infinite cycles (~line 212) when it hit 3 units in a given role before hitting 8 units in a nation (so endless cycling would arise if there were less than 3 roles available). Some of the other crashing I saw was almost certainly also traceable to minimal data usage which had incomplete definitions.

However, none of that should apply with montag_men... unless maybe it was a matter of the system trying to generate descriptions of the recruitable templates? Do those have the same items and graphics defined as a normal unit would? I can't remember exactly what sort of crashes I saw ATM, but my vague recollection was that they seemed like string handling errors of the sort that often arise from description generation...

elmokki:
Montag men indeed had the normal poses as category poses. You should probably make blank "no weapon", "no shield", "no armor" and "no helmet" items and a "no mount" for mounted. Null isn't handled that well in the program.

elmokki:
Also, endless rostergenerator loops are in the "generate units until we have max amount of units" with the failsafe of generating more secondary race troops. If you have `#maxunits 1` poses and only few of  them you definitely need to try to use the commands to limit maximum troop amounts as well.

Ealb:
I used blank items, but still had some array-out-of-bounds crashes on description generation which went away on turning off descriptions.

Agree about where descriptions should come from, ofc. It would be weird not to.

I started with one pose (with all roles) and no maxuints. Stepping through RosterGen in debug, it made 3 units, hit the hardcoded maxvarieties limit of 3 (as passed into generateUnit on line 246), and just kept increasing the max number available for that role indefinitely. It was never going to get out of that units < max loop (as it was >= to the hardcoded maxvarietes = 3, and wasn't changing anything but the limit of units for the pose, which wasn't what was preventing a unit from being made), and it didn't appear to even be considering secondary race troops. Once I duplicated the pose twice and split the copies into infantry, ranged, and mounted, it stopped hanging and completed generation.

elmokki:
The eternal loop has to do with maxvarieties having some hardcoding. Mostly it's limited to weapon amount: It makes sense for normal units to not be able to try to have mroe same armor ID varieties than there are possible weapons.

I really don't think this is a huge issue with foul spawn when they are finished since you should have one unit per pose, but for now I'm adding `#maxvarieties_override` which just sets maximum varieties instead of setting the maximum maximum varieties like `#maxvarieties`

elmokki:
A description crash was caused by units tagged as mounted having a null mount. I made a failsafe for this.

elmokki:
Now, it may still hiccup, but I won't want to dig too deep into fixing it because what you are doing is generating units based on foul spawn outcomes that don't have enough weapons: It's not something the program should handle!

However, now unit with `#maxvarieties` high enough AND enough weapon options should theoretically work, and if not, use the override version. It still seemed laggy and potentially crashing, but at least it  didn't hang up even nearly as reliably.

PR #430, opened 2016-02-28
Mounts without sprites shouldn't cause crashes

elmokki:


Ealb:
Aw, and here I just went and changed my stub mount file to have a blank image.

Also, woo, 1000 commits! ...and it hasn't even been a year. Eek.

PR #431, opened 2016-02-29
Basic fixed-form unarmed foulspawn nations

Ealb:
* Current setup is sufficient to generate unarmed foulspawn with varied stats, costs, and abilities
* Currently since there's no weapons or armor, the almost-functional nations may end up with weaponless foulspawn since free fists are turned off to avoid having both tentacles and fists
* Unit item slots still need figured and added - I'm not sure this will be feasible to do w/o special commands, but I managed to get pretty complicated with hoburg clockworks, so I might be able to figure out how to cleverly manage it with existing tools

### DIFF
--- a/data/items/foulspawn/small/bases.txt
+++ b/data/items/foulspawn/small/bases.txt
@@ -2,30 +2,95 @@
 #name "humanoid"
 #gameid -1
 #sprite /graphics/foulspawn/small/torso/humanoid.png
+#command "#gcost 4"
+#command "#hp 5"
+#command "#size 2"
+#command "#prot 0"
+#command "#mr 8"
+#command "#mor 10"
+#command "#str 11"
+#command "#att 8"
+#command "#def 8"
+#command "#prec 8"
+#command "#enc 1"
+#command "#ap 4"
+#command "#maxage 50"
 #enditem
 
 #newitem
 #name "goblinoid"
 #gameid -1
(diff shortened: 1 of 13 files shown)


Ealb:
How hard would it be to add a set of slot commands/tags/whatever that can be dynamically updated and then define a unit's slots when finalizing them? I.e., the basesprite sets #slot arm 0, the right arm updates it via #slot arm +0 (tentacle, pincer), #slot arm +1 (arm), or #slot arm +2 (pair of arms), the left arm updates it the same way, and when finalizing NG defines the #itemslots command by adding 0 (no arms), 2 (1 arm), 6 (2 arms), 14 (3 arms), or 30 (4 arms) to the appropriate #itemslots numbers for the units heads (1->128, or 2->384), bodies (0 or 1024), feet (0 or 2048), and misc slots (0, 1->4096, 2->12288, 3->28672, 4->61440), with defaults obviously being 2, 1, 1, 1, 2.

Technically I think I could do this via yet more id -1 slot poses coupled with a whole bunch of overlapping #needstype tags (e.g. have a "2 arms" that's #type tworight and #type noleft, as well as another "2 arms" that's #type oneright and #type oneleft), but foulspawn are already getting a lot of slots, and it would be significantly cleaner if running tallies could be maintained for itemslots to avoid the necessity of convoluted pose slot tricks.

elmokki:
Did I understand this correctly? This is just for `#itemslots`, right? If so, it would be easy.

PR #432, opened 2016-03-01
Bronze lance fix

elmokki:
#singleuse does not work. Added #ammo 1 which should work, but I did not bother testing.

### DIFF
--- a/data/items/customitems.txt
+++ b/data/items/customitems.txt
@@ -258,6 +258,7 @@
 #command "#bonus"
 #command "#charge"
 #command "#pierce"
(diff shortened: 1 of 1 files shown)


Ealb:
There's six weapons with #singleuse ([Enchanted] Nets, Copper Feathers, Sacred Pitchers, and [Light] Lances, with the Light Lance being the EA Ermor/Jomon once-per-battle version. I plainly just copied iron Lances verbatim, but I don't know that that tag is actually available to modders, though. In any case, yeah, ammo is how other lances I added work, so that should be fine.

PR #433, opened 2016-03-01
Demonbred one handed flails were... one handed.

elmokki:
Flail is two handed and should be with those stats. Now they're morning stars.


Ealb:
I wanna say 1h flails in the basegame are something you only see on pretenders...

Yup, id347 only shows up on Natarajas and Neters of the Underworld.

PR #435, opened 2016-03-02
Dynamic item slot setting

elmokki:
- Ability to add or remove slots from items/filters/whatever
- Mounted units lose feet slot


Ealb:
This looks like it should give me exactly what I needed to avoid some very complicated pose structures.

PR #444, opened 2016-03-08
Foulspawn stuff

Ealb:
* First pose done, though I'll probably still add hand overlays and possibly "extra limbs" (shoulder tentacles, tails, etc.)
* I ended up using #forceslot for helmets because going the themeinc thisitemslot ... *0 route ended up with #theme elite helmets getting assigned to elite units with oversized heads despite their basechance of 0


elmokki:
I will check what causes the problem with the themeinc way at some point.

elmokki:
Also, I didn't mention earlier, but the approach of adding `#needstype` into weapon slot to generate suitable rarm is probably the best solution to generating stuff like that. Beats the themeincs at least. The themeincs, once they work, are good for cases where you have to determine it the other way around (non-essential slot putting restrictions on an essential one)

elmokki:
Does elite unit in the above mean sacredgenerator elites or troops with slight stat upgrades?

Ealb:
I'm not somewhere where I can check right now, but it's actually possible
it might have only been commanders. I think it was sacredgen elites though.

elmokki:
Commanders sound like a likely candidate that makes sense.

Ealb:
Reviewing the pre-forceslot test mod I'd generated, I did find one example of a sacred wearing an elite helm that didn't fit, but all the other examples were commanders and possibly a hero or two.

elmokki:
I suppose forceslot is good enough. Armor is the primary rescost/prot source anyway. Next time please don't fix it though, I suspect that I know what causes it on commanders but on sacreds I have no idea.

Ealb:
I'm pretty sure I can intentionally replicate this problem on the next FS pose I do. It's probably not that rare if it cropped up in the measly 20-or-so test nations I did before I switched to forceslot.

PR #447, opened 2016-03-09
Foulspawn scouts

Ealb:
* Distinct scout pose for FS; body/limb/head types are strongly weighted (or outright limited) to vaguely humanoid shapes (e.g. no crab heads, but merely a reduced chance of a tentacle)
* This pose kinda looks like it's having a variant of the hat problem; e.g. -749200513 (Ommer) has an assassin with a "tentacloid" head that does #addthemeinc ... not large *0, but is generating with helmet "hood" (#small/#canid) instead of "hood_large" (#large) - this may be different, though, because I have a vague nagging suspicion that I'd previously questioned whether scouts properly process themeincs (in which case it may be considering nothing but the scoutgen boost for #theme "stealthy", which both of those helmets have)


elmokki:
...and excellent place to debug this stuff.

Ealb:
...slightly odd sidebar to these problems. My fork currently has a version (and example seed) where I added (very sloppy) montag template generation to FS, and I found that montag units generated from scout-eligible poses with elite helmet issues appeared to have the pose scout (rather than infantry) despite being a troop pose rather than a commander.

elmokki:
Scout pose has `#role infantry` making it eligible for being generated as an infantry troop.

Ealb:
Yes, but in the .dm file it listed the unit's pose as scout rather than infantry. That may be utterly cosmetic (I don't know that part of the code well enough to even guess without digging through it), but it seemed odd enough to merit mentioning.

elmokki:
It should be utterly cosmetic.

elmokki:
Yeah, can't reproduce it with what I pulled earlier. I had made minor changes but even removing them doesn't change it. I have generated a pile of nations without seeing this bug now.

Also: .#needstype for basesprite, type slave and item basesprite1 on slot offhand failed due to the slot having no items. Roles [ranged], race Dynastic human

elmokki:
Ranged slaves have needstype for offhand that fails due to ranged slaves not having offhands. I could disable the error, but eh.

elmokki:
Okay, now I found the issue I believe.

Hood has `#sprite /graphics/scouts/hood_witch.png`
Hood_large has `#sprite /graphics/scouts/hood_witch.png`

elmokki:
Regarding sacred generator wrong size helmets, multiple helmets are tagged with both #small and #large. Not sure if all of those are intended. At least elitefullhelmet (small, canid, large) looks a bit off on fleshblob (that requires large) because there is a single pixel of flesh popping between the plume and the helmet.

PR #452, opened 2016-03-12
Primarily Yazata/Daeva nerfs

Ealb:
* Yazata/Daeva-only demographic themes for Caelians changed from Y/D-only to Y/D mages/elites with secondary race troops
* Yazata/Daeva are always cap-only and have increased sacred cost multipliers
* +5g Yazata / + 15g Daeva base prices (so 50g/60g)
* non-elite Y/D wings (and thus units) eliminated


Ealb:
I scaled it back from no elites to cap-only elites, but if that's not enough they can certainly be made sacred-only. The more I think about it, the more I'm leaning towards always and only sacred, but let me know what you think.

PR #463, opened 2016-03-21
Foulspawn "cavalry"

Ealb:
cav


elmokki:
I especially low the wyvern and the naga.

Ealb:
The wyvern-lizard-things turned out so well I think I may cobble together
winged lizard mounts. They're *cute*.

PR #464, opened 2016-03-21
Basic data/more graphics for winged FS infantry

Ealb:
* Also winged lizard mount graphics
* And a couple of random missing files/fixes
* If FS are turned back on, the crude FS montag test poses will crash generation every time right now; I'm sure I'm doing something very dumb, but haven't really debugged them yet


elmokki:
At first I got a command problem with the weight argument on something, but then I couldn't reproduce it. Now I noticed how arming units fails for foul spawn templates and that causes looooong (but not infinite) loops in foul spawn generation.

elmokki:
The category poses should definitely not have `#maxvarieties_override 8` since then it tries to generate 8 varieties of same armor from one template, making all the units try having a different weapon. This won't crash it, but it causes a long loop.

elmokki:
Turns out there's also a bug where unit filters can freely set maxvarieties to whatever, making template poses try getting two varieties even with one weapon.

elmokki:
armor_placeholder.txt did not have `#no_armor`, causing some degenerate foul spawn be unable to get armor at all (turtle ones)

elmokki:
`#maxunits` did not work properly. Now it works. It is not max unit amount, it is rather maximum armor varieties while `#maxvarieties` is maximum weapon varieties per armor.

Ealb:
I went ahead and added these explanations to the documentation for the next time I forget what exactly these commands do (and alas, there will be a next time if my past experience proves anything).

elmokki:
These foul spawn nations generate really slow compared to normal ones. No wonder though. I looked a bit at the generation time and big part of it is generating 10 sacreds instead of 1, or hell, even 20 instead of 2 in some cases.

Not a problem though.

Ealb:
I had been wondering where most of the increased time was at; that makes sense.

Is there any other broad style of infantry that makes sense for a pose, or should I move on to crossbred archers and cavalry on crossbred mounts? I suppose I could do an "oversized" infantry pose - not monster-oversized, but size 3 or 4.

elmokki:
There's some horrible looping going on in troopgen too that I've been fixing.

elmokki:
Pygmy only secondary machakans cause eternal loops with foul spawn right now due to their max variety and max troop limits. I kinda recommend changing those for pygmy only nations, but I'm gonna write code around it just in case anyway.

elmokki:
CRITICAL FAILURE, IMAGE FILE /graphics/foulspawn/small/bonus/locust_wings.png CANNOT BE FOUND.

elmokki:
(Gonna fix it myself)

Ealb:
Gyah, typical error brought on by my stubborn refusal to settle on a single consistent naming scheme.

The more I think about it, the more I think I may do a size-3/ogreish FS infantry pose. The idea of cute 'lil ogre-size ettin-things has very quickly grown on me.

elmokki:
That sounds good. I think the minimal releasable foul spawn is fairly close content wise once there's ranged and some special monster style poses up. Then there are some code changes and such required but that's it.

Also, I recommend starting spritegen via Eclipse and just using arrow keys to slide through each slot. Then look whether console has errors.

elmokki:
I am getting what seems like stuck bugs but aren't because separate seeds work when I generate large enough nation counts, which isn't that many. I have no idea what causes that because memory really shouldn't be it as the program can make thousands of ordinary nations.

PR #479, opened 2016-03-31
Ogre armor, ogreish FS

Ealb:
* Fairly complete set of armor for size-3 ogreish humanoids, to include such FS
* Enough limbs, etc. to do a moderately acceptable job of making ogreish size-3 FS

Not there yet, but soon:

mockup


Ealb:
ogres

Still plodding along. Probably going to add bandar/muuch arms (already have heads and legs) and then start writing definitions for size-3s.

elmokki:
They're gorgeous, however the lizard-thing on the first post is the most glorious thing. Honestly beings like that could be a race of their own almost.

Ealb:
They are cute. I did heads for Fire, Frost, and Swamp drakes. Probably could do wyrverns too, come to think of it...

Ealb:
Hmm. Looking at a mockup, there really *is* something to be said for quadruped draconians...

draconian

elmokki:
That might actually suit quite well for some sort of Oriental Dragon (LA Jomon style) race.

Ealb:
Four limbs good, six limbs better?

draconian2

Ealb:
Hmm, I'd need to do a proper torso for them (size-3 FS all have thick fur-clad torsos like ogres), but I'll save my Gimp file where I was screwing around with these for possible future shenanigans.

PR #480, opened 2016-04-06
Last (?) Size-3 FS body graphics for now

Ealb:
* Unless I decide to add wings or random extra tentacles (both of which are admittedly tempting), this should do it for Sz-3 basesprite bits. I'll thus be moving on to defining data and adding any needed equipment sprites (e.g., double helmets for wannabe ettins)
* Also, ehehehe:

bat


elmokki:
That is gorgeous :D

PR #484, opened 2016-04-16
Size-3 Foulspawn (roughly done, still need polish)

Ealb:
* Bigger FS are roughly done and loadable in SpriteGen
* Bigger FS are roughly incorporated into the (montag) poses for Foulspawn nations (generation remains glacial, or possibly became glacial once more - the poses could probably use a looking-over and polishing, as I'm almost certain the template poses could be handled better)
* Generated bigger FS generally look okay, though there are some graphical issues that IMO seriously need fixed. How glaringly awful these will be to someone who isn't me remains to be seen
* Still need to add tails and wings for bigger FS
* Added sea dogs as an Atlantian mount, mostly because I thought it would look good with hunched-over FS (who ironically don't have it, but I'll confess I was thinking more "hybrid" and less "foulspawn" here)

E.g.:

preview_107_napnom


elmokki:
Gorgeous. Especially the weirder ones.

I don't think that much can be done about slow generation speed. I will, however, at some point confirm that it's as optimized as generating equivalent amount of non-foulspawn units.

I also need to fix some things. Probably with fairly simple fixes. Hero generator needs to generate non-category poses and was it so that as PD category poses don't properly get their firstshape? That's an easy fix too at its simplest.

Ealb:
I'm not sure about the PD issue. The only issues I remember with PD were more involved, like weighting poses more heavily for PD (i.e., making FS nations favor category poses over montag poses; you said this would fall into a pile of PD stuff that would eventually require a re-write of that module). In the tests I ran it seemed like it worked properly when the category pose was set as PD; the PD listed the template, and they turned into assorted junk when battle ensued.

If you're just looking to do a couple of quick tweaks, would it be difficult to add a "#commandercommand" tag or suchlike (a less awful tagname would probably be good), so that you could specify commands that are only processed for commanders? Assuming my code-scanning was correct in concluding this currently isn't possible, I mean. The lack of such a command was the excuse I used to avoid going off on a tangent and adding illithid/hybrid (aka FS-lite) nations earlier this week, as starchild-type hybrid commanders all have mindblast, but none of the comparable troops do...

elmokki:
Something like #unittagcommand would probably be great. Then just make all sorts of different generator generated units have suitable tags (even if that means multiple tags in some cases) and allow poses or filters or whatever to affect it too via tags.

PR #490, opened 2016-04-27
Mostly ranged Foulspawn

Ealb:
* Full (albeit narrow) implementation of Foulspawn archers - they caneither be traditional archers, spitters, rare mentalists, or throwers; this variation is assumed to provide some unpredictability as far as quality goes. For the moment, sacreds/elites are disabled, as my data file implementation does not get along with the sacred generator's increased freedom WRT #needstype 
* Foulspawn language themes - these are mandatory, but only control nation names, unit nametypes, and in some cases, homerealm weights. However, with 16 different themes, they should at least have unpredictable names.
* About half of the work to implement starchild mage dynamic robes. This was/is a different sort of headache than the above Foulspawn, so it represents "taking a break" from that, and vice versa (the Foulspawn are definitely the bigger headache. I can't imagine those tangled messes of data files are anything but horrendous to figure out from a layman's perspective.)
* Added some bugs, fixed some bugs

Current state of Foulspawn, more or less:
preview_108_sheneddod
I'd say all we really need for them to be "feature-complete" (as opposed to completely and utterly tricked out to my unreasonable satisfaction) would be to add a minimally complete implementation for mages and monsters


elmokki:
I wrote a long message and accidentally closed the window, but:

- All looks great once again
- Monsters as of now are shapeshifts. Pose based monsters should be easy enough to add as a separate option though. They'd work with basesprite/shadow being given by default and then #generateitem for whatever in whatever order.
- Mages may be some work to implement.

My motivation fluctuates. Monsters I might well do as well as I will fix the foulspawn heroes almost certainly being built from montag template poses at sometime soonish.

Sacred generator and needstype problems might be caused by custom weapons/offhand weapons/bonusweapons not getting tags of their parent weapon and thus ignoring needstypes. It's an easy fix if it is that, but I'll wait for you to verify whether it seems to be that.

I've done some tinkering on ItemGen again if you're interested. It's such light work compared to NationGen since it has to generate items only and it doesn't even need a huge pile of code to be "complete". It's on GitHub but it's undergoing major file format changes that aren't committed because I've been trying to think a good way to figure out paths/costing from code point of view. I have an idea that I have code for, but it's currently untestable as my last changes rendered all the defintions broken.

Here's some of the sprites it currently can do: Picture of items

elmokki:
Also, naming code I will write for either NationGen or ItemGen will for most part be usable for both!

Ealb:
I'll need to have a look to see if the sacreds are generating w/o tags as you suggest. If not, I know an unwelcome but reasonable fix: stop trying to be "clever" and "efficient" and just separate the different kinds of missile troops (archers, throwers, etc.) into different poses with seperate data files. Admittedly, though, when it comes right down to it, "add more data/graphics files" can *technically* solve every problem for NG (except performance, ofc) so long as one doesn't care how many thousands of files a given "solution" might take. I just need to be reasonable about balancing complexity and redundancy. As usual, sigh...

Ealb:
The good side of messing about with ItemGen is that it should indeed be mental relaxation compared to doing Foulspawn (seriously, why on earth didn't I have the sense to actually draft a design document for them? I knew they were gonna be a mess; I suspected *how big* of a mess they'd be...). The bad side is that I still have the attention span of a gnat, and digging through sprite dumps has unfortunate side effects:

forest_giant

(Obviously, that's 10m of sloppy tinkering rather than a well-refined finished product, but it still makes me want to drop everything and make Forest Giants. *siiiiiiiiiiigh*)

PR #502, opened 2016-05-23
Trying to merge net stuff again

### LINKED ISSUE #500: Custom nets do way too much damage
The #dmg for nets is a programming magic number rather than an actual damage value, so if nets get picked to be made magical, they end up doing an obscene amount of damage instead of capturing targets.

Ealb:
* Added #notepic tag support for weapons to SacredGenerator
* Added a safeguard to CustomItemGen that should catch any standard items that should be - but aren't - tagged #notepic because their damage is actually a special effect
* Added #notepic to all nets in the data files, as well as a couple of sticks & stones and rocks, tragically precluding enchanted nets, rocks, and sticks and stones
* This is a re-commit because it appears GitHub is utterly incapable of handling commits whose titles start with "#". Seriously, just never do this. Ugh.
* Or possibly GitHub is borked this morning. Something is painfully wrong with all this in any case.

Closes #500


Ealb:
...

Ealb:
,,,

PR #506, opened 2016-06-03
Giant ichtyid troops

Ealb:
- Just what it says on the tin
- Currently only low-incidence elite/sacred troops; if I add mages, I'll probably make this a theme rather than universal
- So yes, we're definitely currently in the "WTH? Why? Just... why?" phase of the content development cycle

derp


elmokki:
Well, why not?

Issue #505, opened 2016-05-31
A few small things that really need to be done at some point


elmokki:
- Possibility to force sacreds to be cap only
- Possibility to for nation themes to increase cost of priests and mages somehow (Per pick of path? Just plain increase? I don't know. It's for #zombiereanim and such and allows some stronger reanims)

Both are really simple. Also #reclimit on sacred troop daevas is probably in order. I have no idea how high, but probably quite low. I hear they're pretty terrifying.

I may or may not do coding during June and July as I probably will be working way less. The less I need to write SQL to dig through data in SAS, the more I want to code Java. However, I will spend one week of June in Lapland and one in Germany so that takes quite a bit of time. Especially Germany isn't coding friendly.

elmokki:
chaos rec can supposedly return more gold than units cost for cheapest units

elmokki:
you can get full size air elementals for really cheap units (should be tied to unit size for one and increase cost drastically in the first place)

Ealb:
On the last point, Air Elementals of any size are really powerful because they're a free turn-1 flier. I might be tempted to say that limiting any but the smallest elementals - size 2 at the most - to T3 might be reasonable.

For #zombiereanim, increasing priest price is probably a lost cause because if they're too expensive indy priests will be used instead. For #wightreanim and #horsereanim, (or #tombwormreanim, although that's mostly broken; #manikanreanim just plain doesn't work right now) the lower levels don't actually do anything, so indies are mostly not an issue.

Ealb:
I'm not looking at any change in time I can pick away at things, but as usual, my issue is motivation more than time. Well, that and focus. It doesn't help when the only time I spend working on NG stuff is e.g. making ichtyid giants instead of finishing any of the half- or three-quarters-done stuff. (Sadly, true story - that's the only thing even vaguely NG-y I've done in like a week...)

Issue #474, opened 2016-03-25
Foul spawn and hero generator


elmokki:
Currently heroes are category poses. Solution is to add a system that allows #chanceinc generationtag [not] <tag based on module that created the chanceinchandler> and setting category pose probabilities zero except for troopgen and sacredgen and the outcome poses zero for those two and nothing else.

Issue #402, opened 2016-02-15
#price_per_hp and similar commands


elmokki:
I have been planning to do some stuff related to naming and maybe finish the PD generator update at some point (EDIT: Also those nation themes since they'd really help with consistency and allow adding some funkier stuff more reliably), but reading some forums this came to mind. I might even make it a generic `#price_per_command <command> <price>`, but mostly hp scales so well with stuff like regen, glamour etc that those probably need nerfs.

Ealb:
More tools to allow finer-tuned balancing would be nice, yes. Although that of course requires that they get implemented, which *I* certainly am not the best about...

PR #510, opened 2016-06-05
The Great pre-D-day push

### LINKED ISSUE #454: SacredGenerator increasing stat prices
Currently there's no price difference in boosting defence from 10 to 11 and in boosting defence from 16 to 17. This needs to change.

elmokki:
Gotta fix a good pile of issues, both listed and non-listed. Tomorrow is Hearts of Iron 4 release and I will also do some work next week before heading to Germany and then to Lapland so I have no idea when I will work more unless there are urgent bugfixes, which I'll strive to do asap.

This should close #454, #402, #474, #505


Ealb:
Since the PD issue is resolved (even if the engine won't let it be fixed how we'd like it to (AfterTheCat would be absolutely *glorious* if its PD could go full-on montag-random), grr), I'll drag my attention back to FS/centaurs/illithids - even if I probably finish a minimalist-but-acceptable version ogre-kin first, since it should really take long - so we should be able to get a release out in that next-week-ish timeframe.

elmokki:
Yeah. Should also remotely start looking at a 0.7.0 release candidate at some point maybe. 

I got a few things that aren't necessary but I might add if I have some more motivation this week or next Monday:

- Limited terrain recruitment. There should still be support for it. Basically some suitable light troops (forest survival and such) may be terrain recruitable in addition to being fort recruitable. The extra mages could also occasionally be terrain + cap only as long as they don't have astral and blood and have a fairly weak pattern and have either research malus or a cost increase. This should be rareish.

- Stronger foul spawn light/heavy splits. Basically split the category poses a bit and make light/heavy/big/all combinations for infantry and maybe cavalry using themes to make non-suitable equipment rare.

Also, reanimating ranged weapons need a nerf I guess.

PR #519, opened 2016-06-09
Fixed bug with highestcaponlychance

Ealb:
I'm fairly certain this was a bug and not a data error, anyway: #caponlychance is currently used in the data files as a double from 0 - 1, and SacredGenerator was trying to read the arguments to #caponlychance by doing a cdr of the argument, so if the caponlychance was 1 (or 0) it would have an array index out of bounds exception.

### DIFF
--- a/src/nationGen/rostergeneration/SacredGenerator.java
+++ b/src/nationGen/rostergeneration/SacredGenerator.java
@@ -631,13 +631,13 @@ public class SacredGenerator extends TroopGenerator {
 		u.commands.add(new Command("#gcost", "*" + total));
 
 		
-		// The highest caponlychyance for the unit will apply if one is defined, unless the default formula is higher 
+		// The highest caponlychance for the unit will apply if one is defined, unless the default formula is higher 
 		values = Generic.getTagValues(Generic.getAllUnitTags(u), "caponlychance");
 		double highestcaponlychance = 0;
 		for(String str : values)
 		{
-			if(highestcaponlychance < Double.parseDouble(str.substring(1)))
-				highestcaponlychance = Double.parseDouble(str.substring(1));
+			if(highestcaponlychance < Double.parseDouble(str))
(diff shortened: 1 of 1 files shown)


elmokki:
I can't fathom how that piece of code can result in array index out of bounds. Values should contain strings of numeric format: "1", "0.5", "0.33", "0" and so on and the old code would parse those. New code would instead parse "1", "0", "0", "0" and thus result in a bug if the 1 wasn't there.

I also have no idea what cdr is.

Ealb:
Ugh, sorry. "cdr" was me being tired, trying to remember a technical name for "all of a list except the first element", and latching on to the Lisp name for it (and I haven't programmed Lisp for ~17 years, so I have no excuse for this).

I walked through in debug mode when the old code was crashing. The array *str* it was receiving (for a yazata with caponlychance 1 from wings) was a string of the form [1]. There was only the one character in this particular string. *The old code would presumably work for fractional values (I assume Java understands .5 to be the same as 0.5), but it would never work for whole numbers as the whole number would be a string of length one, and *str*.substring(1) would try to read the array starting with the non-existent second character. The new code takes the whole *str*. So to use the example you gave, *values* =[1, 0.5, 0.33, 0] would have previously yielded on each successive iteration (assuming it could keep going instead of crashing): *str*.substring(1) = OutOfBounds, *str*.substring(1) = [.5], *str*.substring(1) = [.33], *str*.substring(1) = OutOfBounds. The new code will yield *str* = [1], *str* = [0.5], *str* = [0.33], *str* = [0].

elmokki:
sounds believeable even if I stiill can't see why it crashed before

Ealb:
If str = [1], str = [1, OutOfBounds, OutOfBounds, OutOfBounds, etc]. substring(1) was always going to start reading with the first of those OutOfBounds if the double was a whole number.

Frankly, I'm surprised I figured this out given how plainly and incoherently tired I was last night.

PR #529, opened 2016-06-10
Chariot stuff

Ealb:
* Added guaranteed "chariot" base name for mounted poses with the "chariot" tag to allow for very-low-incidence chariots for certain unit types where you can't see that they have mounted legs instead of infantry legs
* By that, I *definitely* don't mean pygmies or hoburgs
* And even if I did, it *definitely* wouldn't be because I gave them swan chariots
* Yes, *of course* I should be ashamed of myself, and no, I'm not

sorrynotsorry


elmokki:
These are very cute.

PR #546, opened 2016-06-14
Sidhe stuff

Ealb:
* Stats for fay horse, unicorn, cu sidhe, and giant swan chariots, plus giant swan mounts
* Currently the chariots are relatively expensive (~2x human equivalents) and non-stealthy, but they retain #illusion; it remains to be seen if this is a terrible, terrible idea...


Ealb:
...because obviously this is critical to finishing illithids and centaurs. I blame Larzm for making a 4.25 spritedump (and we'll just forget that I requested it...)

Issue #518, opened 2016-06-08
National spells


elmokki:
Just simple stuff: Get list of current national spells from Mod Inspector, add them or sets of them to two files (a set could be something like "all oriental summons"):

- First file simply slaps everything that doesn't have weight zero or below to the nation. This is for stuff like "Oriental humans should always get oriental spells". There can be variants of the same spell sets just chanceinced suitably that just one is chosen if it's coded properly.

- Second file has a pile of spells that theoretically any nation may get. Plenty of current national spells don't really need to have that specific conditions to be added to a nation. I'd probably have a small non-zero basechance for all these sets in the first place and higher chances if the nation has paths. I don't mind needing a pretender to cast something not worth making a pretender for, that's Dominions.

Also, nations that get a ton of first file spells won't get many second file spells. This is probably also defineable to races or whatever so you can make something weak get a ton of national spells.

Also, it would not be too hard to list a pile of custom national spells as an alternative to the second file. It can be very simply existing spells made better or cheaper or easier to research or something genuinely cool. I for one would love to add stuff like earth-based communion to Agarthans occasionally.

PR #585, opened 2016-07-04
SpellGen code

elmokki:
This is what I wrote in #518 pretty much exactly. It seems to work fine and is enabled now, but it seriously needs more spells boringly plucked from database before it can be enabled on a release (nation.java contains the code that calls generator).

I'll look into making content tomorrow.

### DIFF
--- a/data/spells/custom_spells.txt
+++ b/data/spells/custom_spells.txt
@@ -1,10 +1,10 @@
 --- Earth communion slave
 #new
 #name "Supplicant of Earth"
-#command "#copyspell 'Communion Slave'"
-#command "#name 'Supplicant of Earth'"
-#command "#path 3"
-#command "#descr 'The caster surrenders itself to the earth, connecting through the immeasurable mass through anyone who is willing to listen (communion slave)'"
+#command #copyspell 'Communion Slave'
+#command #name 'Supplicant of Earth'
+#command #path 1 3
+#command #descr 'The caster surrenders itself to the earth, connecting through the immeasurable mass through anyone who is willing to listen (communion slave)'
 #end
 
 --- Earth communion slave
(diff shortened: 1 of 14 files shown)


Ealb:
As a general rule, rather than tying these to races, would it make sense to tie them to tags in the races? I.e., Oriental humans has the #oriental tag, but the oriental theme for Primates adds #oriental to that race, so those would get them as well (and if the Rudhra, etc. spells were tied to #enlightened, they wouldn't get those). In some cases (like Pale Ones or Illithids) where the culture is pretty much inextricably tied to the race, this is right out the window, obviously, and some spell sets would make more sense being tied to fixed races rather than culture-groups.

For reference, the broad cultural theme-groups I've been working with (plus the races that currently defined (or in some cases functionally conceived) as having "representatives" of each culture) are:

* Boreal (Man/Ulm-esque): humans, hoburgs, ogres
* Central (Arco/Sauro/Pan): (Amazon) humans, hoburgs, halfmen
* Imperial (Ermor et al): hoburgs, (dynastic) humans, primates
* Austral (Machaka): humans, hoburgs, primates
* Occidental (Mict/Xib, Nazca(?)): zotz, (muuch) Atlantians, hoburgs, primates
* Oriental (TC/Jomon): humans, hoburgs, primates, (o-bakemono) ogres, ichtyids, tengu
* Dynastic (C'tis): humans, lizards
* Abysian: Abysian, hoburgs, primates
* Fae: sidhe, hoburg, possibly boreal humans, Fomorians (to some degree)
* Nordic (for lack of a better name; I've never made a theme for this): vanir, vaettir
* Enlightened (BL): primates

elmokki:
For most spells yes, this would be better.

PR #586, opened 2016-07-04
Priest level distribution tweaks

elmokki:
Old system was very biased towards H1. Now it's biased more towards H2.

Also a slight increase in amount of stand alone H2 priests even if you
have priest-mages.


Ealb:
Nice. I noticed this long ago and was annoyed by it, but never once thought to mention or change it.

PR #588, opened 2016-07-04
Primary-race tengu

Ealb:


elmokki:
I am extremely afraid of stuff like this, but we'll see. The new sacred generation code may help for one.

Ealb:
In the cold, hard light of day I'm more than a little wary of the way I went about these.By making them elite-only and guaranteeing plenty of secondary race troops, they're assured some cheaper diversity. It might be better to make them more "vannish"; give them non-elite, primary-race-only "tengu peasant" poses which have worse stats but are still overpriced, and reduce their chances of secondary race troops. Basically, do the opposite of what I did last night. That way they may be strong, but they're expensive, fragile, and one-dimensional.

PR #591, opened 2016-07-04
 	Contact Ujigami -> Summon Ujigami

Ealb:
* This leads to a nagot gick fel, depressingly enough.


elmokki:
Gotta release a fix patch after checking that all of the spells match.

PR #611, opened 2016-07-10
0.6.14c

Ealb:
* This is mostly a fix of the archive because 0.6.14b.zip was using some old data files, to include but perhaps not limited to shapechages,txt
* Updated changelog and .jar
* A couple of very minor graphics tweaks for little vinemen and werewolves


elmokki:
Oh, yeah, my bad practice of making binaries from my own branch finally causes a problem it seems since I hadn't remembered to pull your changes to it.

PR #651, opened 2016-07-17
Offset looks odd in SG for winged olm FS 

Ealb:
Slightly odd - the wing overlay is lower than expected, but changing the x offset raises it to the y height that would be expected. It may just be SG being funny because of the recent changes to centauroid-type sprite offsets, but just changing the y offset manually definitely works


elmokki:
Overlays ignore the mount offsets, or leg offsets in case of centaurs. This was originally because they were designed for mounts, who also ignore offsets (and instead force their offsets on others) and easiest way of making them was to crop a piece of the mount. 

I would guess that if it looks weird in SG before you change anything, it looks weird in game. Could you try changing the slot name to something else than overlay and see what happens (renderorder and #needs obviously need changes too then)?

elmokki:
Yeaah. I just disabled the overlays ignoring mount offsets and it works perfectly. What I'll do is add a pose command to do the same and put it to the centauroid pose.

PR #658, opened 2016-07-18
Hoburg thralls for illithids

Ealb:
Because they can't all be impressive


elmokki:
The #needs weapon bite is not a very good practice and my guess is it will be replaced by the claw as the only non-0 basechance weapon in generation.  #needs is generally a good idea mostly for things that aren't generated at all otherwise or that won't be generated again for sure. 

The way troop generator works is:
- Generate unit with only armor (which'll get the claw as #needs is in basesprite)
- Arm copies of that unit

The obvious alternative is to add #chanceinc slottag basesprite this_basesprite_really_needs_claw 1 and set them all 0 otherwise. Alternatively as a less fail prone version keep the claw as 1 and add also #chanceinc slottag not basesprite this_basesprite_really_needs_claw *0.

Ealb:
Ah, yes, so I'm back to my instinctive abuse of #needs. D'oh. In this case, the easiest solution will be to make single-entry fist.txt and bite.txt files. Although... I may instead tag the body slot and use that as a chanceinc trigger to cause the right weapon to be chosen.

PR #665, opened 2016-07-20
Specialized national commanders (slave/animal)

Ealb:
* If a nation has animal troops, it is possible that their commanders will get beastmaster 1-2
* Taskmaster and beastmaster now cost bonus*5g
* #slave commanders always get taskmaster 0-2 (50% chance of 0); if taskmaster > 0, they will also get an inspiration penalty of 0-taskmaster (and get a corresponding gcost reduction of penalty*10g); their overall cost will therefore be modified by -10g (tm +2/ins -2) to +10g (tm +2/ins +0)
* #animal commanders always get beastmaster 0-3 (50% chance of 0); if beastmaster > 0, they will also get an inspiration penalty of 0-beastmaster (and get a corresponding gcost reduction of penalty*10g); their overall cost will therefore be modified by -15g (bm +3/ins -3) to +15g (bm +3/ins +0)
* Essentially, the above allows for non-standard troops to have more/better authority over their own kind but not over others; in the interest of making national commanders appealing, there is no possibility of reduced authority w/o an equal or greater specialized leadership bonus to offset it (and they'll also be cheaper)

### DIFF
--- a/src/nationGen/rostergeneration/CommanderGenerator.java
+++ b/src/nationGen/rostergeneration/CommanderGenerator.java
@@ -108,6 +108,7 @@ public class CommanderGenerator extends TroopGenerator {
 
 		int orig = 1;
 		boolean hasSlaves = false;
+		boolean hasAnimals = false;
 		
 		if(r.nextBoolean())
 		{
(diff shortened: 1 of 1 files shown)


Ealb:
Not 100% sure this won't end up being weird with primate nations; it's possible that instead of having the default non-animal commanders have a 25% chance to get beastmaster if they have animal troops (which will very often be primates) it should only happen if the nation has nature magic. Or else to really screw with things, there should be a "civilized animal" tag (possibly based on the primates' tech level, with it being absent on primitive ones?) on primate races that prevents them from triggering the generation of beastmaster units.

PR #732, opened 2016-10-19
Austral human cavalry (and pygmies) were too fast

Ealb:
* Moved the mapmove 3 entirely out of the race file and added a mounted-specific bases file to ensure they received default mapmove instead of the boosted infantry speed.


Ealb:
In theory, I'm not sure this was necessary. However, I have definitely seen in-game results that make me suspicious that it was (but I'm too lazy to code-dive right now to verify pose/race command precedence, or data-dive to make sure all mounts are defining mapmove.)

PR #754, opened 2016-12-11
Chanceinc command fix

elmokki:


elmokki:
It's worth considering whether the metal men filters should have an hp lower limit as well.

PR #759, opened 2016-12-20
Fir bolg fixes, old magegen fix

Ealb:
* Fixed a bunch of the typical dumb sort of mistakes I make with every new race
* Fixed what was probably a very old bug with racial tertiary mage modifiers trying to use primary mage modifiers and crashing if those weren't defined


Ealb:
Fir bolg nations (barring a fae/more-Nemedian theme or nation-specific mage staves, both of which would be nice but neither of which are crucial) are done and look distinctive (i.e., kinda drab with generally-uncovered red hair). I'll see about getting austral apes finished in the next day or two and then probably push out a patch before the holiday.

PR #768, opened 2017-01-01
Custom item tag fix

Ealb:
* Those were indeed not being copied. There was also something weird going on in Item.java where a list looked like it was being copied onto itself, so I fixed (or possibly "fixed") that as well.
* Fixed heavy sling precision

yoink


elmokki:
I'm unsure where Item.getCopy() is used but yeah, it's the dependencies not being copied that cause the issues.

PR #770, opened 2017-01-05
More dantih armor

Ealb:
* More traditionally "enlightened" armor for tusked ones in contrast with what I already made
* Basically, laying the groundwork for austral and/or central themes (aka content creep)


Ealb:
...I promise not to add a boreal mammothmen theme. I will carefully avoid commenting on the possibility of east-Asian elephants.

Ealb:
...
...
...I'd really like to say that my resolve broke gradually, and that I did this in a moment of weakness, but no, I started messing with this almost as soon as I typed the above. *sigh*

mammoth

ahyangyi:
Mammothmen sound awesome.

Also, by Dominions logic, they should not really be Boreal but just living in mountains (look at Caelum).

Ealb:
The contrast around the belly is something that vexes me - it's hairy, so its figure shouldn't be as evident, but there does need to be *some* more contrast. I picked at it a bit more last night, and I'll pick at it a bit more. It has potential, but yeah, it's not there yet.

Ealb:
Gah, on the subject of contrast, I played a few turns with austral bandar/colossi last night, and the contrast on a lot of their armor looks *a lot* worse in-game than it does in SG, an image viewer, or Gimp. That'll need fixed.

elmokki:
I wonder what causes that. The stretching that generally happens? The contrast compared to background?

Ealb:
I suspect it's the change in backgrounds. That's certainly the most straightforward explanation. I could sorta see that the left torso shadowy area on those units (and some of the legs on the bandar full foo armors) looked a bit iffy, but they looked better than less contrast in those contexts. I suppose it's also possible my judgement is skewed in the short-term by doing most of my work at 800% image size, even when I look at it appropriately sized later; even then, though, the different backgrounds probably are making a difference. I should probably make myself "battlefield" and "recruitment screen" backgrounds to use for eyeballing units-in-development.

PR #771, opened 2017-01-05
Minor updates

elmokki:


Ealb:
Oh wow. That morale bug has got to have been there forever, and I'm slightly appalled that I never checked that area despite from time to time having misgivings about how low some elite morales were...

PR #774, opened 2017-01-06
More sacred morale and magic resistance tweaks

elmokki:
Morale had some stupid code still and I wanted to tweak things a bit to
make them less deterministic.


Ealb:
Hmm, looking over some vanilla nations, it might not be bad to give heavy troops and cavalry a base +1 morale, or a decent chance at such. Right now there's sort of an invisible tier between line troops and elites. I suppose this could be handled in poses or by adding a not-uncommon "veteran" filter, though.

elmokki:
It should be done in troop generation so that units of same type get same bonuses and it could be done with a free filter, but since it'd be very universal it's probably better to put it in code.

PR #788, opened 2017-01-15
Reclimit for dustwalker chariots

elmokki:
I heard these are pretty op. This is a very soft nerf.


Ealb:
Seems right. Enc 0 on a trampler - especially one that can be sacred - is potentially quite nasty.

PR #789, opened 2017-01-19
Balance changes and bugfixes

elmokki:


Ealb:
Pffft, roll them back? That goes against the current NG division-of-labor plan, where I giddily make unbalanced nonsense and you act as the voice of reason and reign me in slightly. Speaking of which, I'll try not to make too much work for you in this regard with the full-sized illithids I'm about to add, but I can't promise anything since my current "but how can I ever convince people to recruit armored mindflayers" "solution" perforce leans towards, uh, untested waters...

elmokki:
I'm not great with balance either considering it's years since I last played the game properly. I just listen to people complain and try to do something.

Ealb:
Well, to make matters worse, while I play a couple of MP games per year, the metagame I play in is comparatively cautious *vis a vis* an awful lot of metas. *sigh* At least we both have very strong understandings of the mechanics...

PR #791, opened 2017-01-24
Big squid

Ealb:
* Full implementation of illithids
* I ended up taking a conservative tack with the illithid soldiers, though I may revisit the issue; right now they're size 3 that eats as 4 (I may wanna go with ressize 4 too, but we shall see) and formation fighters, with either a breastplate and helmet or a robe and some sort of defensive perk that jacks up their gold cost... they're probably still not gonna be desirable


Ealb:
For the record: I'm not sure I'm actually happy with how the nautilus head for illithid mages turned out (no matter how logically it follows from the squid heads), so it may get axed. It's... I dunno. I have very mixed feelings.

basesprite

PR #795, opened 2017-01-24
0.6.21b

elmokki:


Ealb:
The obnoxious thing with this bug is that Advanced Descriptions would report reasonable prices for the basically-free mages, which is probably why we never heard about this with #warriormage...

elmokki:
Yeah, I very rarely launch the game which is probably why this ran under my radar as well.

PR #801, opened 2017-01-26
Excluding some weapons from further upgrade

Ealb:
(Mostly but not only poison slings)


elmokki:
I actually wonder whether it was necessary for poison sling or whether there's a bug in code.

elmokki:
...right, so, there's #dt_poison and #poison. First makes the weapon do damage in ticks and second makes it genuinely poison. Anything that gets #dt_poison should probably get #poison too even if database only has a field for the former.

That does, however, also remind me that the damage type changing magic weapon effects may need to check for existing damage types. I'm not entirely sure how well for example some of them work together.

PR #810, opened 2017-02-12
Hoburg fixes and #additionalweaponslot

elmokki:
...this can be used for four armed things if it is expanded via cleanUnit() to all units instead of troops!


Ealb:
Oh geez, those were awful data definitions. I mean, your observations are correct, but the definitions were jacked up even taking that into account.

However. Looking at hoburg clockworks again (and having an attention span that makes mayflies jealous), this seems like a good time to follow the lead of Avvite nations with their hero-only giant poses, and take the next ***ahem*** logical step in re: the hoburgs' cultural Napoleon complex:

clockwork_evolution

PR #815, opened 2017-02-16
Optimizing nation generation and code cleanup

Freykin:
I'd ignore the nationgen.jar in here, as it's not up to date.


elmokki:
I'm gonna raise an issue with this that is not really necessary to do but should make code less cumbersome to work with if more filters are every added.

PR #821, opened 2017-02-19
Periodic changelog update

Ealb:
(This isn't suggesting we need to do a release soon or anything; I'm just trying to be proactive.)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,30 @@
+0.6.22 (???th of ??? 2017) - current through 19 Feb 17
+----------------------------------------------------------
+Code
+----------------------------------------------------------
+- A huge pile of refactoring and optimization
+- Added handling for #specifiedgcost and #specifieddescr to monster definitions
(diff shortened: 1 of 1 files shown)


elmokki:
There's no real problem in releasing soon, though the nation filtering still needs to be tweaked before we can release.

It's better to release often when there's nothing blatantly in progress.

Ealb:
Aye, I was mostly thinking about filtering when I said that. Although ofc I
also have countless in-progress things that I "should" finish *right now*...

On Feb 19, 2017 4:05 PM, "elmokki" <notifications@github.com> wrote:

There's no real problem in releasing soon, though the nation filtering
still needs to be tweaked before we can release.

It's better to release often when there's nothing blatantly in progress.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/elmokki/nationgen/pull/821#issuecomment-280948938>, or mute
the thread

PR #825, opened 2017-03-01
Mostly naga work

Ealb:


Ealb:
Also, tightly-coiled naga basesprites for chariot poses so they can tool around with most of their slithery bits hidden inside chariots...

PR #826, opened 2017-03-02
Many small fixes

Aciho:
Mostly these are typos, or incorrect mod commands. Also removed a few commands that don't appear to exist. Two slightly more major changes:

1) Removed #mind, made mindless check for #mor 50 instead

2) Stopped adding #att, #len and #dmg to armors. It looks like the armor generation stuff is a work in progress, so I just bodged it a bit, I imagine it'll get fixed properly later.

It looks like I may have converted the line endings in some files, so sorry about that. I believe github deals with that in a fairly nice way so hopefully it won't be a problem.


elmokki:
Holy shit there are many tiny bugs! Thanks for these, really.

Ealb:
Eek, a lot of these are because of my bad habit of using the ModDB as a reference in place of the modding manual. There was certainly a lot of non-functional slop in any case. Thanks, yes.

Aciho:
No problem. I realise most of this stuff is impossible to test in any sensible way given dom4's relaxed attitude to error checking, so fiddling with the mod inspector gave me a perfect opportunity. I would do more but I'm not much of a Java dev.
I did find one other bug I couldn't track down, I'll add it as an issue when I'm on my other PC and I can reference it properly.

PR #827, opened 2017-03-06
Faster swan chariots, some flag graphics

Ealb:


elmokki:
I really don't think 6 and 7 are fast enough. I'd just put it to 12 or so and claim that the swan can fly while pulling the chariot.

PR #834, opened 2017-03-23
Work

Ealb:
* More graphics for dantih/tusked ones/elephant people
* Graphics for olms
* Cleaned up pale one files a bit
* Implementation of ancient olmspawn
* Implementation of ancients pale ones with steel gear
* Graphics for granite/onyx pale one gear
* Some ancient pale one graphics were tweaked/revised/Dom4'd


Ealb:
704e213 looks weird when I review it in certain contexts. I'd appreciate it if you could check if you get a chance to make sure I didn't commit an out-of-date version along with my changes. I'll probably check tomorrow night when I'm more awake than I am now (i.e., when I trust my judgement) if you don't get a chance to look at it.

For the record, the factor that made this an abnormally complicated commit was that two of the files I was committing had more recent versions in the repository (customitems.txt and zotz.txt), and I used Githib's inline commit conflict resolution (which is new, I think? I hadn't seen it before) to fix that. Looking in the repository, it seems like the commit was correctly executed. Looking at some diffs in some contexts (but not others), it looks like it may or may not have been. Ugh.

If there's any doubt, just revert the commit and I'll make sure my branch is free of conflicts rather than trusting in new, complicated Github features.

elmokki:
I have no idea why it looks so weird, but to me it seems that it includes a pile of commits that already were in main branch as new changes. For example, it shows the lines I changed in MageGenerator recently as a new thing.

So, no problems I guess.

Ealb:
Looking back, it seems like it may have taken everything that wasn't up-to-date in my fork, merged it *into* my fork, and then reported the process of merging into my fork as though it had been a merge into the master. There's a messy sort of logic to that, I guess.

PR #839, opened 2017-04-05
Reduced/increased upkeep filters

Ealb:
Filters using the new #addupkeep command to directly manipulate effective costs for upkeep:
* Added T2/T3 power 1 filter that effectively reduces upkeep 40g/year (non-sacred)
* Added T2/T3 power -1 filter that effectively increases upkeep 40g/year (non-sacred)
* Added elite filter for gcost 15-30 elites that effectively reduces upkeep 8g/year and gives +1 mor
* Added elite filter for gcost 30+ elites that effectively reduces upkeep 16g/year and gives +1 mor
* Added sacred filter for gcost 15-30 elites that effectively reduces upkeep 6g/year and gives #neednoteat
* Added sacred filter for gcost 30+ elites that effectively reduces upkeep 12g/year and gives #neednoteat

### DIFF
--- a/data/filters/default_elitefilters.txt
+++ b/data/filters/default_elitefilters.txt
@@ -107,6 +107,23 @@
 #end
 
 #new
+#name "loyal-10"
+#basechance 0.25
+#chanceinc personalcommand gcost below 15 *0
+#chanceinc personalcommand gcost above 30 *0
+#command "#addupkeep -10"
+#command "#mor +1"
+#end
+
+#new
+#name "loyal-20"
+#basechance 0.25
+#chanceinc personalcommand gcost below 30 *0
+#command "#addupkeep -20"
+#command "#mor +1"
+#end
+
+#new
 #name "resistances"
 #basechance 0.0
 #chanceinc "magic air 0.1"
--- a/data/filters/default_magefilters.txt
+++ b/data/filters/default_magefilters.txt
@@ -1,4 +1,24 @@
 #new magefilter
+#name "highupkeep"
+#basechance 1
+#command "#addupkeep +50"
+#notfortier 1
+#power -1
+#theme upkeep
+#desc "Greedy"
+#end
+
+#new magefilter
+#name "lowupkeep"
+#basechance 1
+#command "#addupkeep -50"
+#notfortier 1
+#power 1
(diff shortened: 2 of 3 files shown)


Ealb:
The mage filters might be too much - I went with +/- 50 for upkeep calculations (+/-3.3g per turn), but that seems like it might be a lot for power +/-1 - I know I have a bad track record with estimating what would be good in that regard.

elmokki:
There's probably a semi-hidden command modifier like %gcost20% or something that is used for something and that may or may not work. That might be good here if it exists and works.

elmokki:
This is what I was talking about (and I'm not 100% sure the thing you implemented works but I suppose you tested):

#new
#name "chaosrec10"
#type "chaosrec"
#basechance 0.05
#command "#chaosrec %cost10"
#chanceinc "command #chaosrec 50"
#chanceinc "command #chaospower 50"
#theme recruitment
#power 2
#end

Ealb:
Ha, my search terms were too narrow when I looked for what you meant.
Actually, I haven't tested that mess yet, but it's how mage healing (and nothing else) is handled, and I walked through the relevant chunk of code when I turned it up. I'll probably look at this and my mess tonight; the scaling-by-cost rather than -by-tier looks like the better choice, though.

Ealb:
Ugh, the %cost thing is a bit of a mess. It's redundantly processing the "#addupkeepcost %cost50" repeatedly - the command gets added to percentCostCommands an absurd number of times (#chaosrec does as well), which seems like it's usually at least 100 but I've seen as high as 1000. If the command uses %cost50 it still works because it'll overwrite that #addupkeepcost every time, but if it's the awkward-looking %cost-50, you end up with messes ranging from e.g. -3000 to -12000 because it subtracts the value from the running total. Moving the population of percentCostCommands to polish() fixes most of that, but there's still a few issues with how handleCommand deals with old arguments starting with %. When those are fixed, it's pretty much spot on, though getGoldCost() is returning too-high values for sacreds (at a glance, it looks like the *1.3 sacred cost mod might be getting applied to finalized prices rather than non-finalized prices).

elmokki:
I am not entirely surprised there's something wrong there.

PR #842, opened 2017-04-08
"Bloodguard" elite filter

Ealb:
* Added blood-nation (moreso human, less so size 4+, never for undead, demons, or magic beings) elite-only filters modelled on Bloodguard: boosted stats and berserk with no cost increase in exchange for a reclimit and homesick.
* Added 14 guarenteed name prefixes for them
* Added a prone-to-affliction nationwide template as a mirror of "Hardy"
* Tinkered with occidental animal warrior armor graphics a little


Ealb:
TODO: go through and add "general race tags" to pretty much all races for ease of broad, sweeping chanceincs (mostly just humans, as those are the main race spread across multiple NG races)

elmokki:
If I ever end up making more contextual naming things like this need to happen.

Like, as I've probably described before, instead of just having a bunch of words assigned to fire magic for example, make the program realize that words like lava and volcano are connected and that if one is used, the other is a good word to use as well. Similarly it'd be neat if, for example, a three mage setup themed would get somehow themed names so that instead of random path-specific words they'd be somewhat connected. There are definitely better examples, but for example we could have a general mage theme of "nature" and then "forest" is a sub-word for nature for a nature mage while an astral mage gets "moon" or "stars" or something instead of "secrets" because that's connected to nature as well.

Ealb:
If I were more ambitious than I currently am, I'd dredge up my half-forgotten graduate studies and try to make some kind of semantic web (or use an existing one, though the first one that pops into my mind (WordNet) suggests an extant one would probably need a lot of wrapping to be useful to us) to do this sort of thing. Mind you, I'm still interested in doing something like that, and if we start a re-work I'd geek out over that faster than I would pretty much anything else related to NG (as all that mage name crap I did while originally getting sucked into the project probably made clear), but I don't have the motivation to start doing anything like that right this moment.

PR #843, opened 2017-04-08
Actual bloodguard filters

Ealb:
I managed to avoid uploading these with the other stuff, d'oh.


Ealb:
Ha, and the low version of the filter has an extraneous chanceinc that I didn't notice until I uploaded it. I'll fix that whenever I next commit something.

PR #851, opened 2017-06-20
Montag PD support

elmokki:
Sadly this did not work for start armies, but oh well, the PD is the most important part.

### DIFF
--- a/src/nationGen/nation/Nation.java
+++ b/src/nationGen/nation/Nation.java
@@ -1034,15 +1034,15 @@ public class Nation {
         
       
         tw.println("");
-        tw.println("#defcom1 " + pds.getPDCommander(1).id);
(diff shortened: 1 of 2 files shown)


Ealb:
Does that mean that start armies will be a template unit on turn 1 and you won't see what they're gonna be until turn 2?

elmokki:
Yes. Also, I just realized this is NOT how it used to work so I will need to separate start army and PD generation.

Ealb:
Also, looking over your branch update pull, I am suddenly extremely appalled I didn't draw a line (or several lines) under things and push out a patch (or more) this spring. Blah. My target for wrapping things up on content will be naga, dantih (so we have an "excess of elephants" theme going), and finishing integrating the flag stuff AlStar contributed (because other people's contributions are more important to prioritize than mine, as far as I'm concerned).

PR #857, opened 2017-07-09
Mostly flags

Ealb:
* A bunch of flag top graphics; almost all of these should "work", although there seems to be something off about their alignment vis a vis existing flag tops (those seem like they're too far right by a pixel) 
* A couple of flag centers; these need to be tested on generated flags and given a long, hard looking-at before they're deemed "good enough"
* A couple of variant naga colors, based on pretender snake patterns; not sure what if anything to do with these
* "Ageless pale ones" (these are the Earth Made Flesh with gills and cloaks added) for a size-6 hero-only pose, or maybe also as a sacred slow-to-recruit cap-only monster option for healthy pale one cultures...


Ealb:
The intent of this commit is mostly to finish off the flag diversity work (well, graphics-wise, anyway) already in the 0.6.22 candidate. The random other stuff is my short attention span at work.

PR #876, opened 2017-07-24
Camel mounts

Ealb:
* Camels as mounts for base humans (and by transitivity of the mount file, amazons, advanced humans, foulspawn, and oriental humans) that have hot national temp preferences, austral humans, humanbred, avvim, illithid hybrids, and enlightened/imperial/oriental vanara with hot climates.
* I considered giving these to hoburgs, and I may yet relent, but for now no hoburg camelry
* Updated name files to accomodate camelry
* Graphics for celestial hounds as an oriental-only mount; too lazy to do the definition right now


elmokki:
This reminds me that perhaps some mounts should be made rarer in certain climates. Bears and wolves at least live in fairly warm environments from what I know, but as mounts or war beasts they might fit better to nations without heat preferences. I cannot think of any fantasy shenanigan of bear- or wolf-riding people from specifically warm climates. Warhammer fantasy goblins are probably closest I can think of.

Ealb:
It somewhat randomly occurred to me that the venerable, haven't-thought-about-it-in-20-years Elfquest would fit the bill for high-profile hot-climate fantasy wolfriding. It's not common, though. Keeping them for eastern cultures but not anything west of India might be reasonable; there's certainly enough Asian wolves and bears in jungles for that to be justified, especially if we're talking vanara moreso than humans on them (and smaller bears/non-dire wolves rather than the bigger ones).

PR #878, opened 2017-07-27
Misc graphics

Ealb:
* Bandar thralls (all demographics) for illithid nations; I'm not entirely happy with these, but I've been picking at them on and off since December, so I just made myself finish them
* Ichtyid cavefish thralls
* Some more weird foul mounts/monsters I may or may not do something with


elmokki:
I think these are great. However, the elephantoid is the greatest.

Ealb:
The original plan was to scale it down and replace a bear or horse head with it, but then I noticed that a decapitated elephant was just the right size and almost exactly the right color to take the tentacles. And so...

Which of the boaring grubs looks better, BTW? I'm torn between the simple-but-cleaner head swap and the busier-but-less-cliche head-belly-tail look.

elmokki:
I'd just put both in.

Ealb:
...I will admit I'm also tempted to use the bandar thrall graphic as an uncivilized ape mount for pygmies...

PR #904, opened 2018-02-03
Restrict race/theme BEFORE random selection

EMcG0402429:
A couple fairly easy changes to speed up nation generation and unfuck the algorithm a bit.

Reduces the list of primary races randomly selected from to ensure that it includes only the Must Have Primary Race restriction list, and does not include any of the Must Not Have Primary Race restriction list.

Ensures that where Must Have Theme has at least one selection, the first theme selected is one of the required. (This is still properly filtered, and may still cause nation rejection if, i.e. If you have Races of [Abyssian, Muuch, Atlantean] and Themes of [abyssian_themes - primal, muuch_themes - default], you won't get any Atlantean nations out of that, and will still see a bunch of rejected nations.

### DIFF
--- a/src/nationGen/nation/Nation.java
+++ b/src/nationGen/nation/Nation.java
@@ -197,7 +197,7 @@ public class Nation {
 		addRaceThemes(races.get(0));
 		
 		// Add nation themes
-		addNationThemes();
+		addNationThemes(restrictions);
 
 		// Secondary race after themes since themes may affect it
 		allRaces.clear();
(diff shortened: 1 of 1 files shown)


EMcG0402429:
Also I don't actually know Java, so this may have problems with it which were not immediately apparent to me.

elmokki:
While your code does make sense, it has a big flaw for the program: If seed 123, for example, normally produced a nation of humans, under any restrictions that do not allow humans the seed would now produce something else. This is not something I am willing to allow since I want all seeds under given settings integer and version to produce the same nation.

Things would be like this already if it was easy. However, the only way to really make it work the way I would accept it would be to bake in the restrictions into the seed somehow. This is definitely possible, but a lot more work.

EMcG0402429:
Ah. Knew I had to have been missing something. Will think on this further then.

PR #907, opened 2018-02-04
Some more flag icons

Ealb:
Ported some low-hanging-fruit as far as flag icons go. The serpent was literally the small serpent icon resized, so I did the same with the bone serpent and tentacle. Ur's ziqurrat was also just scaled up, so I did likewise with the pyramid, though I did smooth that one.  The plumage icons were based on fairly large pretender graphics, so they work all right even without scaling up; same with the horror. The large number of plumage icons is based on an assumption that there is no color masking.


elmokki:
You can use color masking if you wish.

Ealb:
Well, I'm not sure it's really needed for these. The plumages were all colored roughly according to path colors, and hand-coloring them, even only roughly, does improve the shading a bit.

elmokki:
There is, by the way, alternative recolorer code that is better but not turned on since it makes some stuff I manually brightened for the old code too bright. There's some command to activate it for specific things that I should probably find.

elmokki:
If enough stuff is just greyscaled making that default might be best.

Ealb:
You've mentioned it before, and I promptly forgot it.

elmokki:
The alternative system is fairly old and simply uses a more sensible way to colorize. If you rename Drawing.createColorizeOp and Drawing.createColorizeOp_alt it should be 100% functional.

Basically for a greyscale image the default one sets each color as grey/255 * color, which darkens the image needlessly: A 128/0/0 red on 128/128/128 greyscale becomes 64/0/0 red, much darker than the image that originally was greyscaled. 

The newer one is more complex, but essentially it makes the above color combination result in 128 red while bright white results in bright red. The only potential issue is that super bright colors might get clipped: Bright red recolor will result in a very bright red image. 

Now, the important part is how a simple greyscaling in GIMP is reproduced when image is colored back: Bright red becomes 127/127/127 grey with lightness desaturation and much darker with the other options. 127/0/0 red becomes 63/63/63. Thus while theoretically things might become super bright, a bright red robe is saturated into 127/127/127 grey and thus even bright red does not clip the color channel and reduce detail.

The problem is the original human robes used by a pile of nations that I brightened for the older coloring system. Had I had any experience in the math behind recoloring the alternate system would be the default one, but I only made it later when I stumbled upon on how it actually should be done.

PR #915, opened 2018-02-27
Generalized recpoints, Caelians

Ealb:
* Changed the default unit values to include recpoint 10000 to match the default gcost 10; there should no longer be any units with rpcost 0 being generated, though many fancy races like elves will have obscene recpoint costs
* More rigid start age for different Atlantian sizes; this will need expanded and repeated for other races where age corresponds to size like muuch and pale ones
* Did recpoints and mapmove for Caelians; significantly, reclimits are removed from daevas in favor of a 30-point recpoint penalty (with Adorable Ones getting a 20-point hit). Both also got a 5g price reduction.  Under Order 3, that translates to ~5 per round in the cap, but that's with a penalty of -3 to the stats affected by chaos power. Under Turmoil 3, it was ~2 per round. This *looks* like it should be a safe and more organic alternative to the hard limits, though of course it'll need proper testing by people more inclined to abuse things.
* Fixed some dumb typos


elmokki:
I was thinking about simply setting #recpoints = 10000*(sum gcost from race, pose and basesprite), which should be mostly fine seeing how for example vanir have 25 rec points and EA Van footmen are 25 gold. This would be, of course, only set if there wasn't a #recpoints command already existing.

Then to adjust for stat changes that autocalc won't there can be stuff like #adjustrecpoints +1 for each stat increase of sacred generator and #adjustrecpoints -1 or #adjustrecpoints *0.8 for unit filters and so on.

Ealb:
As long as there's a basepose #rpcost set, the autocalc seems to do *reasonably* well just based off price... although I really want to decouple those a bit, so having some commands in NG to tweak them would be very helpful.

On an unrelated note: 5.15 gives us canonical sheep sprites. ~Ram~ Mouflon riders, here we come!

elmokki:
The one thing that makes #rpcost require coding is the militia style filters. They really need to reduce rpcost and reducing it by 1 or 2 on a giant is a bit stupid. Everything else could be done by just adding #rpcost +1000 or +2000 as commands.

I have no ETA when I can be arsed to do this though.

elmokki:
Also do note that we're still using Dominions 4 databases. There's code to automate a large part of the database update if you download the correct files from Dom5Inspector, but there's also annoying busywork of adjusting the labels to match because original Dom3DB did not for some reason have commands as labels and the program evolved to use those labels instead of commands when querying database.

elmokki:
...and all that the databases matter for, mostly, is that if you use Dom5 weapons or armor, some chanceincs will not trigger properly because there's no database info on that and sacred weapons might be absolutely fucked up.

Mentioning sacred weapons, I wonder how the alternative sprite system for them is working. Badly, probably, because as far as I remember the alternative sprites spawned under extremely strict conditions. That's something that might warrant a fix eventually, though honestly with the new blesses I don't think flooding the sacreds with secondary effect weapons is a good idea anyway.

Ealb:
Hmm, Jotun Miltia have the same rec point cost as other Jotun (30k) but since their gcost is only 20g that probably works they end up at 11rp vs. 17rp for regulars. Likewise Fomorian militia, but they have REALLY cheap recpoints for their ages (30rp, but the non-elite size 4s are 15g militia, 20g LInfx2, and just one 30g warrior - so 7, 11x2, 17). It kinda seems like the vanilla solution is massively reduced gcost for giant militia, which does make sense mechanically - a huge unit with garbage stats is a punching bag. Like, uh, pale ones - who have 12rp (!) militia vs. 18rp (!!) regulars.

elmokki:
I suppose we need more militia filters if we want to fix this properly. Or conditional effects, but I'm not making those as they are essentially the same.

Ealb:
...here's a question worth asking. Should mechanical mounts/exoskeletons be high res/low RP? It kinda seems like that would make sense; there's just the one pilot, and even if it's taking artisians time to forge the things, those workers aren't part of the normal pool of potential recruits - and by recruits, I mean both hoburgs and horses...

(Ugh, there's a thought that rhymes with "fiddly file editing". It's arguable that exotic mounts that are currently expensive could cost markedly less, but have a significant RP hit to represent breeding the wyverns or whatnot...)

PR #922, opened 2018-03-11
Rec point work

Ealb:


Ealb:
Are we merging stuff into NG/Dom5 or NG/master? #918 (your most recent I think? It was the one with automated body types) went into master, but all the other updates either of us have done have been into Dom5. 

I also need to remember to look at shapes in addition to mounts and monsters in re: mapmove and such.

elmokki:
Merging to the master branch was a mistake, but then again I see no real reason to have a separate Dom5 branch in the first place so might just as well merge everything to master and stop using Dom5.

PR #926, opened 2018-03-11
Changelog

elmokki:


elmokki:
We should probably release as soon as you feel we're ready.

Ealb:
I should be able to do the MM/RP changes for mounts, monsters, and shapes tonight.

As far as actual release vs. release candidates, what else needs done? Fort defenders?

elmokki:
Testing at least!

Fort defenders I did not even think about but yes, they probably need implementation.

Ealb:
Heh, yes, testing is definitely important. I'm close to 100% certain there's at least 1 basesprite/race/pose that has a gcost of 10000+ from a cut/paste error assigning RP costs, despite my best efforts to check and re-check...

Ealb:
This has ended up taking longer than anticipated, specifically in the shapes piece regarding interactions between body types and copystats. It may just be that the best course of action is to never assign a body type if there's a copystat command present, but I'm too tired to think things through cleanly and finish this right now. Hopefully I should be able to get things done enough for a release tonight, though.

elmokki:
I am quite sure #copystats is not an issue if it comes first, though there's no reason to use body type if the command exists I suppose. The automatic thing skips the automation if there is #copystats present too.

There's obviously no hurry. I have no idea when I can be arsed to code in the fort defenders. Hopefully soon.

Ealb:
Ha, so what I was worried about was already there? If I was that tired it's probably just as well I didn't keep working on it.

elmokki:
No, the code for the automation doesn't work for shapeshifts right now, but I suppose I could make it do that.

Ealb:
I can just stick body types on the things that don't have copyspr, but it would probably be safer to have it in place if it wouldn't be too much trouble.

elmokki:
Okay, I thought I was done with this, but it turns out I just generated a mod with an unit (a slave) that costs 9000 gold for some reason.

Ealb:
Probably my fault. This would be the kind of cut-n-paste error I was alluding to above with #gcost being pasted and not changed to #rpcost despite the number being changed. What race?

elmokki:
Yeah, metic slave bowmen have gcost 9 and gcost 9000. I'm gonna fix it.

PR #934, opened 2018-03-17
Enough stuff for RC2

Ealb:
* Mapmoves for all mounts
* MM/RP for all monsters
* MM/RP for all Foulspawn (these may be wildly unbalanced as I tried to
be clever; each FS base gives what amounts to a RP "pool", and depending
on heads/arms/legs they may be well under or very over what the RP cost
of the base allows. However, this is irrelevant until/unless there is
code added to calc avg montag RP costs...
* All sacred bonus wings that I could think of have +6 MM
* All the templates that affected MM that I could think of have been
updated


Ealb:
This should be enough to throw out there for other people to find the sneaky errors, so feel free to do so (I'm not feeling awake enough to trust myself to do it at this point). I did minimal testing and found only one small (now-fixed) problem. Well, Foulspawn still need something done for their template poses' RP costs. But other than that, most everything we were looking to pull together for this RC is done.

The RP numbers seem a bit high, but that's to be expected, and it'll probably take some massaging to get those to behave well. I'd also like to play with some slightly less harsh pricing for RP-expensive units, but that's a later thing to do...

Ealb:
(To clarify, FS montag units right now are REALLY RP expensive. But they're playable. Just disadvantaged. It'll need fixed eventually, but it's not enough to merit holding back the RC unless you want to do a really quick fix, which admittedly would probably just involve averaging the RP costs.)

elmokki:
The montag rp costs are a bit of an issue since we're just setting the base cost for any unit. Getting any reasonable automated average would really require figuring out how the game autocalcs the costs.

Ealb:
Hmm, I can slop one out. I may try doing a hybrid system where I try to guess what a "reasonable" baseline gcost for a pose type is, and assign that x1000, so if a montag's average gcost is higher that particular montag costs more RP, and vice-versa. My temptation is to err on the side of generosity, because it seems like Foulspawn should be fecund.

PR #943, opened 2018-03-20
Bugs related to nation and race themes

### LINKED ISSUE #942: Climate preferences from themes don't appear to be applied
Right now, it looks like any nation which has a theme entry of the form:

#racedefinition "#nationcommand '#idealcold -3'"

is not having any climate preference set; those who have this set directly in the race file still seem to be getting it. This kinda worries me because if this isn't being set, there may be other themes-modifying-race-definitions shenanigans that aren't being applied. I've had no time to test this beyond noting that it's happening, and I'll not have a chance to look at this before tomorrow, so I don't know if it's more severe or something dumb and simple.

elmokki:
- Anything giving +X to a command nation did not have was handled wrong.
- Nations did not take nation commands of races from race themes into account at all.

This closes #942

### DIFF
--- a/src/nationGen/entities/Race.java
+++ b/src/nationGen/entities/Race.java
@@ -125,7 +125,6 @@ public class Race extends Filter {
 		
 		tags.removeAll(toBeRemoved);
 
-		
 		this.handleOwnCommand(str);
 		
 	}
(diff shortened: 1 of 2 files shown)


Ealb:
FWIW, I'd say you should do RC3 whenever you think things are ready. I'm gonna be traveling over the next week, so I'll not have access to a proper computer nor as much free time, and likely won't be touching things to any meaningful degree if at all.

PR #955, opened 2018-04-09
Filter fix

Ealb:
Removed/simplified conditional chanceincs from mage filters, as there
appears to be a problem with them, or possibly just with personalmagic
and them. I'm not alert enough tonight to try to figure out if the
problem is conditionals, personalmagic, or my use of conditionals, but I
may revisit examining this when I'm more able.

### DIFF
--- a/data/filters/default_magefilters.txt
+++ b/data/filters/default_magefilters.txt
@@ -519,12 +519,12 @@
 #name "sun awe"
 #basechance 0.1
 #command "#sunawe +1"
-#chanceinc personalmagic astral OR personalmagic fire 0.4
+#chanceinc personalmagic astral 0.4
+#chanceinc personalmagic fire 0.4
 #chanceinc personalmagic astral fire *2
 #chanceinc personalmagic holy *2
 #chanceinc personalmagic death -0.5
 #chanceinc personalmagic blood -0.5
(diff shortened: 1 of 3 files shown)


elmokki:
How was it broken? Did it trigger when it shouldn't or not trigger when it should or both?

I'll see if I can be arsed to set up a better testing environment later. It's really something I should've done ages ago. Like, generate a nation from a seed and then run whatever chanceincs on it or its units. 

Personalmagic is pretty complicated since it has built in AND-condition and level checks, but something like #chanceinc personalmagic astral OR personalmagic fire 0.4	should definitely be fine.

elmokki:
It's probably fixed now.

Ealb:
Sorry, I should have been a whole lot clearer. It was crashing, actually.
I'll re-check a bit later today to see if it's fixed.

Ealb:
It looks like this is fixed. Among the other things I should have mentioned in my tired, sorry excuse for a bug report was that when I went through in debug mode to the point of crashing (which was checking for the path on a unit, well past the point when the chanceinc should have been split), the chanceinc was listed as "personalmagic astral OR personalmagic fire 0.4", which very much points to the whole not-treating-uppercase-conditionals-as-conditionals being the problem.

PR #959, opened 2018-04-11
Tusked one Kami-of-the-Moon-style kimonos

Ealb:
(I'll probably try to tie things up into RC4 so we can get some more bug reports tomorrow. Probably. If I can avoid doing more Extremely Important Things like making kimonos for elephantman wizards.)


elmokki:
Also, I'm not going to do anything, but one thing that might be worth considering if you can get motivation for it is to re-do some of the older races. I mean, the basic humans for example aren't very pretty.

Ealb:
...giant hoburg god-kings looming over their tiny, teeming terrified slaves...

Base humans could use a makeover, yeah. Even just re-shading the basesprite would probably help. When I tweaked the pants to be more Dom4-ish, it helped, but the bodies and some of the armor would benefit from some massaging. That seems like the kind of thing I might do randomly sometime when I'm bored, when I'm doing something with them and their look annoys me, or sadly more likely, when I really should be doing something else. I'll make an issue so I don't totally forget it, though.

PR #971, opened 2018-04-21
Dantih, done

Ealb:
* Robes for dantih casters
* A bunch of stupid bugs fixed based based on generating a couple hundred nations
* A whole bunch of graphics tweaks based on generating a couple hundred nations
* I'm not 100% happy with how they look - e.g. the falchion pretty much touching their head looks off to me, but not enough to prompt me to change it - but I'm *fairly* happy - and when it works, it works pretty well IMO
* I still could use to add demographic themes so that mammoth men are possible, but this is good enough for now. Likewise Austral regional themes for African-elephantmen, or if I were ambitious enough to do all the spriting, elephentmen in samurai armor, hehe.
* Definitions pass SpriteChecker, but I've not done in-game testing
* Probably completely unbalanced


Ealb:
We could probably do a RC this weekend - there's not been a ton of changes, but there were likely enough. If you feel like doing one tomorrow that'd be keen (I'll be occupied all day), but o/w I'll likely do one Sunday.

What's our 0.7.0 criteria, BTW? It seems like we should be close given how much of the stuff needed to get optimally Dom5 compatible is done vs. what isn't.

elmokki:
Do an RC absolutely whenever you feel like.

I think the 0.7.0 criteria should be that there's nothing blatantly broken and it's mostly Dom5 compatible. Feel free to release that too when you feel like it. I really have no idea what's missing.

elmokki:
In an ideal world, of course, we'd have an RC out for some time and then release that RC + bugfixes as 0.7.0 rather than the current continuous development stream (which could still happen, just as a separate branch to be merged after 0.7.0). While this would maybe be nice for people who want to start MP games or something, I don't think it's necessarily worth it. I mean, people who let players choose their own nations too freely are calling for a disaster anyway.

Ealb:
That, and since NG is a metamod, we could completely and utterly break a subsequent release w/o having any effect on existing games, so that particular justification for clean/dev splits seems a bit less pressing.

I'm pretty sure we have the core Dom5 functions done. There's some edge stuff that'd be nice (possibly doing something with spellsinger/spellsinger communion stuff, or if nothing else playing with the names to make sure only #spellsingers get named Spellsinger; maybe fix the mess with Agarthan cap-only pact spells; maybe do something obscure but incredibly important that I totally had forgotten...), but I think we've hit all the major points.

Well, come to thing of it I did kinda wanna shift some of the cost of sacred custom magic weapons from gcost to rcost. Right now that's a nasty double whammy since it makes them cost more gold and RPs, and in vanilla examples it seems more likely to be factored in through rcost anyway.

PR #981, opened 2018-04-29
Nonstandard fort support

Ealb:
* Added advanced late era fort level to default_forts
* Added [fortified villages->wooden forts] to default_forts
* Added giant_forts, city_forts, and ice_forts; various races/themes were assigned these in place of/in addition to default_forts
* Updated Advanced Descriptions to indicate if a nation builds giant/city/ice/village forts (default forts produce no special output)

### DIFF
--- a/data/misc/forts.txt
+++ b/data/misc/forts.txt
@@ -4,7 +4,7 @@
 #new
 #name "0: Palisades only"
 #type "era 1"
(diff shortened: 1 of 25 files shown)


Ealb:
This might be the last Dom5 thing really missing, but I was thinking the same thing a week ago...

PR #982, opened 2018-05-01
Fixed #chanceinc race <value>

Ealb:
* #chanceinc race <value> was checking whether <value> == "race"

### DIFF
--- a/src/nationGen/misc/ChanceIncHandler.java
+++ b/src/nationGen/misc/ChanceIncHandler.java
@@ -2296,7 +2296,7 @@ public class ChanceIncHandler {
 		}
 		else if(args.get(0).equals("race") && args.size() > 1)
 		{
-			if(u.race.name.toLowerCase().equals(args.get(args.size() - 2).toLowerCase()))
+			if(u.race.name.toLowerCase().equals(lastarg))
 			{
 				return true  != realnot;
 			}


elmokki:
There might well be more of those that slipped by. It used to be so that the program passed the whole chanceinc into that function and back then last argument was the modifier.

Ealb:
Before I go and do a bunch of changes, is there a reason why scouts seem to be picking hands before basesprites? Order is hands, shadow, basesprite, legs to make a naked unit, then on to weapons and the rest of ScoutGen. My modular spy/assassins almost work now that they actually look at race when picking basesprites, but their hands are random colors right now. I can easily fix it by either frontloading race selection into hands rather than basesprite, or by renaming the hand slot to something else entirely, but it seems weird and I don't remember having this problem with any other type of poses.

Ealb:
Hmm, actually, I'm not sure scout basesprite selection accounts for #needs based on the stepthroughs I did to debug the race chanceinc, so I may just need to rename the hands slot skin or somesuch.

elmokki:
It's not ScoutGen specifically but rather unitgen.generateUnit() which is used for many things. I suppose the reason is that #needs hands <foo> on basesprite would be broken if the order was switched. Other than that I don't know if it matters.

Honestly though, I would imagine that a modular scout setup would have separate poses for each race with hands/legs/basesprite/shadow/wings being distinct racial files. That should avoid random hands in the first place. 

There is, however, a very easy way to bypass this if another order is for some reason desirable: Skip the slot if the unit already has an item in the slot. This is probably a desirable thing to do so that #needs for different parts of the unit don't have to be in specific items for them to not be overwritten.

Ealb:
The odd thing is that I was doing #needs for hands on basesprites, and that never seemed to fire - it just picked one randomly, then went on to pick a basesprite afterwards.

Ealb:
(And OMG, changing that from "hands" to "skin" broke things strangely - basesprites no longer generated.)

elmokki:
> The odd thing is that I was doing #needs for hands on basesprites, and that never seemed to fire - it just picked one randomly, then went on to pick a basesprite afterwards.

Yes. Currently the way it works is:

1. it generates hands and uses #needs to generate basesprite
2. it generates shadows
3. it generates basesprite and replaces the one #needs put there originally

It'd really be best to make it not generate anything if slot isn't empty since it'd solve this kinds of fairly strange behavior unless you know how the code precisely works. After that the order doesn't matter (for those three slots anyway)

elmokki:
Oh wait, no, I read it wrong.

elmokki:
It should, by all logic, work right now if basesprite has "#needs hands foo". 

ScoutGen calls UnitGen.generateUnit() that generates first hands, then shadow, then basesprite. The generation happens by running chanceincs, so it does result in null if all the options have zero probability after adjustments, which might somehow cause your empty basesprites.

The new item is set by Unit.setSlot() which does handle #needs on the new item, provided of course it isn't null and the #needs command isn't bogus, though Unit.handleDependency() does have multiple System.out.println()-calls that would warn unless you have written #needs wrong in which case it doesn't even try to run it. The exception is that if you use #needstype, it does not give an error message if it cannot find a suitable item.

elmokki:
The reason for your slot name changing (though please don't do it if you can avoid it) basesprite to null at some point is probably related to dependencies somehow.

Ealb:
I'm doing something very wrong, because I can no longer replicate the results I had right after I fixed that chanceinc (when I was getting hands that seemed randomly picked, but with basesprites that were not missing - the only thing wrong was the basesprite's #needs hands foo wasn't doing anything). After doing then (by hand) reverting the hands->skin slot change thing I did, I'm no longer getting basesprites to generate, just like before I fixed the chanceinc. I'm going to take a break from this, and maybe I'll be able to figure out what horribly stupid nonsense I did later. I'll probably actually upload my current mess now, just so I have a backup of where things stand... like I should have done earlier when everything was almost working, sigh.

PR #998, opened 2018-06-19
Make Seeds 64 Bits

flash-fire:
With my last change, I noticed a user in discord running nearly a million iterations.  That's 0.02% of the seed space itself. Unacceptable.

Anyways, change the type of the seed from Integer to Long. While, this technically appears to not change the result of a given nation, it will change stuff if you use a fixed seed to generate an assortment of nations.

NOTE: I haven't tested the console. But, I've tested the GUI fairly extensively, IMO.

### DIFF
--- a/src/nationGen/GUI/CommandLine.java
+++ b/src/nationGen/GUI/CommandLine.java
@@ -10,9 +10,9 @@ import nationGen.NationGen;
 public class CommandLine {
 	public CommandLine (String[] args) throws Exception
 	{
-		Integer seed = null;
+		Long seed = null;
 		int amount = 0;
-		List<Integer> seeds = new ArrayList<Integer>();
+		List<Long> seeds = new ArrayList<Long>();
 		int settings = 0;
 		boolean nextseed = false;
 		boolean nextamount = false;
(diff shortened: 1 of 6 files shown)


elmokki:
There's a conflict here after merging the rest. I am too lazy to open Eclipse to fix it myself.

elmokki:
I sent you a collaborator invite so you can just fix this and merge whatever else at your own pace. Though if you change something fundamental very drastically, letting us check it out first would naturally be nice.

flash-fire:
Thanks!  And yeah, I'll avoid merging less simple PRs without closer review.  For now, I'm honestly only comfortable working with things that won't change seed compatibility.

elmokki:
Breaking seed compatibility is nothing to be afraid: They'll break any time anything in the actual generation is changed in any manner anyway.

The thing to be afraid of is making something that makes seeds non-reproducible. These tend to be mistakes in reworking (or adding) things that are handled in NationGen.java that relate to nations. Currently this would be at least:

- Nation restrictions. Adding new ones is not a problem, but reworking the system could be.
- Non base game items. The custom items nationgen uses for generic troops are handled there so that ids can be reused. Magic weapons for sacreds are also stored there even though they honestly don't need to since they're not reused between nations. This, at least the magic weapon part, should probably be cleaned up, but it's not too easy to mess it up if adding features generating more magic weapons as long as you follow things done in CustomItemGen.getMagicItem() (mostly the end of it)
- Some handling of shapeshifts that probably could happen in Nation.java is done in NationGen.jar too

The real reason not to touch generation is that large parts of it are horrible spaghetti. I also think that instead of Nation.java calling different generators it should be NationGen.java calling generators with the nation as an argument, but these are the types of rewrites that are boring to make.

PR #1002, opened 2018-07-11
Refactor custom items into separate class

flash-fire:
Most of the code was just shifted so it's honestly a fairly simple refactoring. 

For regressions, my testing is generating a set of 5 seeded nations & comparing the output file.

### DIFF
--- a/src/nationGen/CustomItemsHandler.java
+++ b/src/nationGen/CustomItemsHandler.java
@@ -1,0 +1,194 @@
+package nationGen;
+
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.elmokki.Dom3DB;
+
+import nationGen.items.CustomItem;
+
+/**
+ * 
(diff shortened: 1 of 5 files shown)


elmokki:
That looks simpler and neater than I expected such refactor to be! I can't see why there'd be any problems.

The secondaryeffect loop is there to handle custom secondary effects. There seem to be two of them, old CBM style boulders that I don't know if they're even used and mechanical mosquito bite's specific fatigue effect: #secondaryeffect mechmosquito_fatigue

However, the if(isNumeric) is just pointless indeed. The else-part should be if(!isNumeric).

But yeah, thanks. This is very nice.

elmokki:
Also, nations.csv is probably unused. I'm pretty sure it's a relic from when disabling vanilla nations needed manually setting each of them to a different era, and even then it was a bit stupid as the code could've set all nation ids below the first one used by the program to a different era.

PR #1012, opened 2018-08-10
Misc Refactors & UI Improvements

flash-fire:


flash-fire:
@elmokki Is it okay to remove the default_nationfilter since it has been unused since nation theme were added?  Otherwise, I'll merge this.

elmokki:
It is fine. I am unsure why it was kept in the first place.

PR #1041, opened 2019-01-24
Dom4->Dom5.  zombiereanim became priestreanim

flash-fire:
Was pointed out this bug in a new game that's started-- morrosnationgen-- on the main dom discord.

### DIFF
--- a/data/themes/default_nationthemes.txt
+++ b/data/themes/default_nationthemes.txt
@@ -159,10 +159,10 @@
 #end
 
 #new
-#name "zombiereanim"
+#name "priestreanim"
 #type "reanim"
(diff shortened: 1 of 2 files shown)


Ealb:
TBH, this should probably no longer exist as a national theme in its current form. This is pretty much "like Scelaria" as far as intent goes, and from Dom4 to Dom5 Scelaria went from (to put it in Dom5 syntax) "nation has #priestreanim" to "every national priest has  #reanimpriest". Right now in vanilla, only domkill nations get this version of reanimation; i.e., the ability to use unholy mobs of indy priests for reanimation.

PR #1102, opened 2020-04-05
Enforce entity data definition integrity

lord-brimstone:
I can't believe some of these entity definitions that were completely unused because of a missed #new, and I can't believe it was just ignoring botched definitions until now... haha

### DIFF
--- a/data/items/abysian/humanbred/lightarmor.txt
+++ b/data/items/abysian/humanbred/lightarmor.txt
@@ -20,7 +20,6 @@
 #enditem
 
 #newitem -- full leather
-#newitem
 #name "full leather"
 #gameid 15
 #sprite /graphics/abysian/armor/leather_hauberk.png
--- a/data/items/advancedhuman/normal/armor.txt
+++ b/data/items/advancedhuman/normal/armor.txt
@@ -18,7 +18,6 @@
 
 
 #newitem -- full leather
-#newitem
 #name "full leather"
 #gameid 15
 #sprite /graphics/advancedhuman/armor/leather_hauberk.png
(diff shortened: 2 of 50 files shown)


elmokki:
I would not be surprised if the fact that the program allowed #new when an object was already open and just ignored commands if #new hadn't been specified is an example of how the program was written: If something throws an error.

I mean, the fact that there was ``else if (instance != null && strLine.startsWith("#"))`` for handling normal commands does seem to me that I used it to fix an error.

In retrospect, it would've made a lot of sense to go for JSON or something.

lord-brimstone:
It's true, there's a spectacular amount of error-swallowing in the code, and I've been trying to move away from that in favor of "failing fast" so that we can actually figure out what's wrong with either the code or data definitions.  This commit and one of my huge pull requests last year have found quite a few typos and things that were pretty easy to fix!  

The next thing I want to target for integrity enforcement is the fact that any unrecognized commands fall back to "okay, this must be a tag, just add it as a tag"... which does make the syntax nice and short for adding custom tags, but it also completely hides typos in non-tag commands.  There's also the issue that some things that are tags probably shouldn't be tags and should be full-fledged commands, because the code relies on certain tags, and by nature of tags, it can't detect typos with them.  I might force the `#tag "mytag 37"` syntax so I can throw an exception on unrecognized commands, but there are like thousands of places to convert, so I'm holding off for now.  Hmm, typing this out, I just had an idea for defining tag definitions in a separate file.  Define the name of the tag and what type of args it allows, and that way we can detect unrecognized or malformed tags and commands while still also allowing new tags to be added without changing the Java code by just making a definition for the tag in the data.  Hmm... maybe I'll assign myself an issue to do this.

And I hear you on the JSON, but I do also appreciate matching the style of Dominions mod files.  Though one of my wishes is for Dominions mod files to be written in an actual scripting language...

PR #1105, opened 2020-04-13
Elder cyclops demographic themes

Ealb:
* Poses, etc. for elder cyclops troops (rare and few) and mages (more common)
* Themes dictating when and how these show up vs. lesser cyclopes
* Graphics and definitions for Indean-style dhotis for bandar and dynastic/dustwalker mages


Ealb:
I'm not sure about how some of the helmets look - it's possible the crested open-faced helms should be eliminated.

PR #1110, opened 2020-04-23
Removing diacritics in Nordic nation names

Ealb:
* Dominions is Nagot gic fel-ing when Nordic nations have diacritics in their names, so all name parts with letters featuring diacritics have had the letters replaced with uninflected versions of the same letter


Ealb:
As Lord Brimstone has fixed this, I'm going to go ahead and revert this.

PR #1115, opened 2020-04-26
Translated characters with diacritics for filenames

lord-brimstone:


lord-brimstone:
I can't check an actual run of the code for another 3 hours when I'm off work, but I just checked the code where it should be happening, and it all looks good to me.  

Are you positive you've merged my code into yours?

Ealb:
Yep - and I'll admit the code looks like it should be doing what it says it should. The one thing that concerns me is this:

`	private String sanitizeForFilenames(String in) {
		return in.toLowerCase()
			.replaceAll(" ", "_")
			.replaceAll("Ã¥", "a")
			.replaceAll("Ã¤", "a")
			.replaceAll("Ã¶", "o");
	}`

That doesn't look right as far as the characters being replaced goes, for fairly obvious reasons, which makes me wonder if this could be a localization issue on my machine.

Ealb:
Ugh, it looks like it might be. Manually replacing "Ã¶" with "ö" generates a filename of oresheim instead of öresheim.

lord-brimstone:
Ooh.  Yeah.  Definitely an encoding issue.  On GitHub the characters display properly.  See if you can change the file encoding in your IDE to UTF-8

Ealb:
Ugh, yep, it was encoding as Windows-1252. Sorry to trouble you with this...

Ealb:
(I now wonder if the file sanitation was even necessary...)

lord-brimstone:
I had tested it on my end before my change and it was also giving me errors, so the change was necessary.  We should probably also report it as a bug to Illwinter, though -- they probably also have some sort of encoding issue specifically when their code looks at file names.

PR #1131, opened 2020-05-11
Custom earth communion bugfix, crab balance

Ealb:
* Added missing parens to Supplicant of Earth definition
* Line troop crabs lose AP on their pincers, +5g/+5rp, elites/coms +5g/+10rp


lord-brimstone:
I'll try to add a check for extra command args -- instead of just grabbing the first arg and being done with it, I can make it yell about any extra dangling args.  That should catch the issue in the future.

Issue #1137, opened 2020-05-22
Items use #natt but the command is #nratt


djmcgill:
The items that nationgen generates don't properly get the number of attacks that is intended due to the typo. PR coming up very shortly (git takes 30s+ for each checkout in this repo!).

PR #1138, opened 2020-05-22
Fix #natt and #ambidexterous typos

djmcgill:
I think a little bit of whitespace autoformatting snuck in too

Closes https://github.com/elmokki/nationgen/issues/1137

### DIFF
--- a/data/items/customitems.txt
+++ b/data/items/customitems.txt
@@ -483,7 +483,7 @@
 #command "#magicarmor"
 #command "#ironarmor"
 #end
- 
+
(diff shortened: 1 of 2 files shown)


djmcgill:
Actually while I'm at it the command is "#ambidextrous" not "#ambidexterous" I'll fix that too one sec

djmcgill:
There we go. I did notice BTW that some units generated with #ambidexterous AND #ambidextrous so I don't know if there's some stuff going on with the base template vs additional ones.

PR #1153, opened 2020-06-25
Minor bugfixes to site paths and unit descriptions

geepope:


src/nationGen/misc/Site.java
```diff
@@ -35,11 +35,12 @@
...
 		
 		List<String> lines = new ArrayList<>();
-		
+		MagicPath path = getPath();
+
 		lines.add("#newsite " + id);
 		lines.add("#level " + level);
 		lines.add("#rarity 5");
-		lines.add("#path " + getPath());
+		lines.add("#path " + path.PathToInteger(path.name));
```
lord-brimstone, line 43: This can just be `path.number` (or `getPath().number` if you wanna remove that variable you created)

src/nationGen/naming/NationDescriber.java
```diff
@@ -52,14 +52,6 @@ private void describeUnits(DescriptionReplacer dr, List<Unit> units, Filter desc
...
 				
 			
-			for(Filter f : u.appliedFilters)
-			{
-				if(f != null)
-				{
-					desc.append(f.tags.getString("description").map(s -> s + " ").orElse(""));
-				}
-			}
-
```
lord-brimstone, line 62: Can you describe what this change achieves?  I'm not really sure what the purpose of this method is

geepope: This is the method that generates the description for individual troops. This loop in particular appends a brief description for each of the current unit's filters (e.g. "Poison resistant", "Strong", etc.) This same loop shows up identically earlier in the method, with nothing between the two loops that would modify what they do, serving no apparent purpose other than duplicating the filter descriptions. This results in unit descriptions such as "The infantry of foo are armed with various weapons and light armor. Poison resistant Strong Poison resistant Strong", which gets even worse when you get into filters that have longer descriptions. This change removes the redundant loop to make troop descriptions a little more readable. (This only affects troop descriptions. Commander descriptions are set in a separate method which does not feature the redundant loop.)

src/nationGen/misc/Site.java
geepope: Thanks. Sorry, still working on getting my bearings.

src/nationGen/naming/NationDescriber.java
lord-brimstone: Gotcha, I see it

PR #1156, opened 2020-07-08
Caster leadership cleanup and tweaks + Rimvaettir poses and themes

geepope:
* Fixed and cleaned up leadership calculation for mages/priests

New method randomizeLeadership replaces largely redundant if/else trees in generateMages and generatePriests. There are some minor nuances of the old generation that are not captured but for the most part it's a fairly accurate approximation of what the old method would have generated, hugely simplified and condensed. Could probably be expanded to encompass mundane commanders.

Originally, extra mages did not have their normal or special leadership set (and certain mages had special leadership set twice, probably intended for extra mages.) This consolidates leadership calls and includes extra mages.

* Cleaned up and condensed extra mage chances

Consolidated a lot of conditionals for extra mage chances and removed a redundant boolean. Preserved existing logic for all extra mage chances, except as noted there was an extra 1 in 10 for nations without guaranteed earth 1/fire 1 only - this seemed very arbitrary for such a small change that I just took out the restriction.

* Cleaned up ApplyFilter

Took out a redundant loop in applyFilters related to tier selection.

* Added a more or less complete set of rimvaettir poses as a subset of the vaetti race

Could possibly be fleshed out a little bit but has most of the stuff from vaetti plus some additional magic gear. Graphics could be cleaned up and used as a base for the new-style vaettir but for now it's kind of nice to have the visual distinction between the two sets of vaettir.

Additionally, regular vaettir poses had a renderorder issue that frequently stopped offhand weapons from being visible that is now fixed.


lord-brimstone:
Looks okay to me.  Did you run the data integrity tests?

geepope:
Yep, the integrity tests were super handy throughout the process. Still tests OK as of last commit.

Ealb:
I've looked over this stuff, and there's a few tweaks I'm looking to make
(making crossbow poses non-pose-based, tweaking some theme numbers (mostly
ice modifiers), and some tiny nitpicky garbage for consistency, to the
degree that NG themes are consistent, and also some graphics tweaks like
bonus weapon y-offset). It looks good in general, though, and as long it's
passing the diagnostics it should be good on a deeper level than my lazy
superficial eyeballing.

Ealb:
Hmm, some of the mage robes need tweaked/excluded for rimvaettir as well.

geepope:
> Hmm, some of the mage robes need tweaked/excluded for rimvaettir as well.

What did you have in mind? I mostly made this as an experiment to see how much work races were, so I'm not 100% sure on inclusion criteria. Were you thinking thematic reasons or are there graphical problems I'm not seeing? (Not trying to be defensive, just curious.)

Ealb:
You'd be warranted to be a bit defensive - you did a good job with a pretty messy addition, and I am a notoriously error-prone contributor. That said, doing these wretched files is a lot like coding, insofar as it's very easy for mistakes to *look* right because you know what it's supposed to do. A second set of eyes is something that I find to be very helpful, and which could have prevented numerous errors I've introduced myself.

There are a few different issues - I'll make an excessively verbose version of what really should be a far-shorter list and those we agree should change you can fix, or I will.

The #subraceprefix tag needs to be on the rimvaetti bases, not the themes (and I'd go with rimvaetti rather than rimvaettir for the subrace prefix, as the singular form seems more correct for adjectives in mostly-English unit names).  Right now, all vaettir will be labeled (Vaetti) in advanced descriptions even if they should be (Vaetti - rimvaetti), and you'll never have an e.g. Utborgian Rimvaetti Swordsman - it'd just be an Utborgian Swordsman.

I'd be inclined to make rimvaettir themes generally a bit less common overall - right now they're present to some degree in ~40% of all vaetti nations. I'd also be inclined to weight them more heavily towards EA - even though vanilla recruitable ones are LA, it's more consistent with general vanilla trends for things to get less magical over time. Pure rimvaetti nations should be much more probable EA and rare if ever LA, etc.; by LA it should mostly be no more than lingering remnants in most vaetti nations.

Some of the theme numbers don't quite fit, either. Some of this is very nitpicky; e.g., there's no reason for tiger to be 0.25 compared to wolf/boar/bear/etc, as Siberian tigers are very much a thing. Some of the ice numbers in tribal themes seem a bit odd as well; spider should be quite low, while deer (and again, tiger) should do okay with it. It might just be better to remove ice from the tribal themes entirely and leave that to be determined by demographics (although since there's currently no ice armor/weapons, it's not a big deal). On a related note, it'd probably be best to use demographic themes to limit tribal themes; e.g., temperate/warmer-weather themes probably shouldn't be available for nations with any rimvaettir.

Overall, I think I probably need to do a pass over the #magicpriority tags in vaetti themes - I'm 80% sure they haven't been updated since the way this tag works was changed. These were the first or second set of themes I did and it kinda shows. Still, having said that, if any demographic themes in a race are doing #magicpriority tags, it's usually best to remove any ones in the race file and have each demo theme explicitly define their own.

Right now, rimvaettir-only nations that have the advanced tech theme will have crossbowvaettir. That shouldn't happen. This can be fixed by removing the vaetticrossbows pose set and adding its pose to vaettitroops, gated by "#chanceinc NOT racetheme advanced *0"  Pure rimvaetti nations can also have monsters that give them vaettir riding on moose-/mammoth-back; that needs excluded (or replaced with rimvaetti variants).

Graphically, I noticed that harpoons, etc. are too low on rimvaettir's backs by several pixels. 

Robe-wise, loincloth_fur is... odd. everything above the belt should probably go; on the flesh-and-blood vaetti everything above the waist is a semi-transparent overlay adding a very subtle pair of breasts to the vaetti's naked chest. I see no need for ice goblin hags to have mammaries, so everything above the waist (except gold chains for the ornate loincloth version) could probably go, to include the odd white dot on the arm.

Aesthetically, the bulky bear/deer/etc robes could probably use a bit more contrast, as could the Spenta robes (and to a lesser degree, the druid ones), but that's not a huge issue.

I'd also be tempted to add a subtle recolormask to the trim on the leg armors. Re: legs in general, vaettir are an old race design and they're doing things in a way I try to avoid these days; if I were to re-do them or add something as expansive as you did I'd eliminate the legs slot, add naked-ish legs to the basesprite, and make armors cover the whole body. This would mean adding a mounted basesprite and full mounted armors instead of normal armors plus mounted legs, but even if this adds a bit of graphical redundancy it generally makes for data definitions that are *a lot* cleaner.

geepope:
Cheers, thanks for the detailed explanation. I was having trouble figuring out what was going on with loincloth_fur, I must have messed up the transparency when copying it over to rimvaettir and wasn't sure what I was looking at. Axing it is probably the way to go.

I defer to your thoughts on thematic considerations as I was pretty much just arbitrarily throwing numbers out. The only sticking point is that I would think it would be a shame if Vaettiheim mixes were too rare in LA, but I agree 100% that the rimvaettir-heavy mixes should skew heavily EA. (I implemented troop poses before caster poses so I left age considerations out of the first demographics I put in so that you'd have a reasonable chance at Vaettiheim in LA, then forgot to go back in and add age modifiers when I put in full rimvaettir demos in.)

(Also, you make a comment about there not being ice armor/weapons currently that I'm a bit confused by - they're there, but only the rimvaettir elite poses get access to them. The ice modifiers are already functional as far as I can tell.)

Ealb:
Ha, I somehow managed to forget I saw them earlier in the week, and I must have been only looking at a normal troop when I synced my dev computer's copy and tried to recreate my list of possible issues. As previously mentioned, I have a solid reputation for being error-prone, and it's not at all undeserved.

Re: Rim chances, w/o looking my instinct is to say make everything above "mixed" twice as likely EA, and half as likely LA; make mixed half as likely LA; make elites only twice as likely LA and all-vaetti half as likely EA and twice as likely LA. MA chances might need more attention than I'm giving them here, though...

D'ya wanna handle tweaking stuff, or ya want me to do a polishing pass? It's no problem for me to do it, especially since it'll let me work on NG w/o finishing my own unfinished work (*siiiiiiigh* - I do intend to try to get things to the point where a release is feasible this weekend, though)

geepope:
If you don't mind taking a polishing pass, please go ahead (particularly on the art front, since as you say it's tough to match; I had a hell of a time with the contrast on color masks, as you noted.) I started this to take a break from staring at the code but now I could use a break from taking a break at staring at the code...

elmokki:
There's a thing with the recolor masks: The original code I wrote way back was stupid and required some adjustments to recolormasks. I rewrote it later and added some command to use it instead. The  new code should produce okay results by just greyscaling what you want recolored.

Ealb:
A year or three ago you told me about your newer better recolormask process - I wanna say in a comment on an issue? Possibly one of the old open ones? I'm not sure now though. You explained the command *somewhere* on Github though, even if I can't remember where to find it...

elmokki:
It took a while to find thanks to all the (very sensible) refactors, but Drawable has this:


```
			if(i.tags.containsName("alternaterecolor"))
				colorizeFilter =  Drawing.createColorizeOp_alt(c);
			else
				colorizeFilter = Drawing.createColorizeOp(c);
```

So, if the item has #alternaterecolor, it probably uses the alternative recoloring function.

I would've liked to force the alternate to be default, but a lot of content might have required graphical reworks I for one was not interested in doing. It may be worth testing how things do by swapping which one is the default.

PR #1181, opened 2021-01-15
Austral Shenanigans

JAM-564:
This is an overhaul of austral themes, focused almost entirely on pygmies and making them, if not competitive, maybe capable of defeating easy ai.


mandarbmax:
Seems like a good idea, pygmies aren't very good as it stands currently. What about Colossi Pygmies as represented by giant austral hobrugs? It might also be fun to give the pygmies either totems like Vaettir have or move them to other regions like how you can get roman monkeys.

JAM-564:
I kinda think that pygmy only should be a barely playable meme. I want to believe that the strength of the australs is in their amount of choices, not spastic themeing but just a variety of sizes and good iron gear. Especially if I can see if I have the ambition to set up Colossi into a more playable state, then austral human might be a decent contender against the currently size 3 dominated meta. 

The drake pelt colossi could bring nut cracking weapons to smash apart a heay armour size 3 nation while not taking too many resources themselves. The pygmy mages have strong patterns, so they could support there. The machakans I view as the best option for a frontline, what with long weapons, heavy armour, and an actual hp pool. And the pygmies have a role in being extremley powerful siege strength/defense if they get the right filter or if fire and astral magic is online, they can either flaming arrows anything to death or take an average of 96 hits per square that is luck/ethereal buffed. 

I dont think they need more themes, just a greater variety options and more bells and whistles to the current advanced/iron/primitive and maybe a better pose selection for their culture themes.

Ealb:
One thing with something like pygmies is to remember that races can be both secondary races and have secondary races. Trying to make every race strong enough to stand on its own as a contender ignores that, and can lead to power creep.

Not every NG-generated nation is competitive. Changing it so that every one is would require either a whole lot of balancing code or a whole lot of homogenization. I'd rather see pure pygmy themes get their chances increased when secondary and/or have higher secondary race troop ratios than to make them extremely strong on their own.

PR #1182, opened 2021-01-17
Further Austral Rework

JAM-564:
This aims to make australs more different. Mostly making Colossi to be an aid to pygmies and machakans, just as pygmies exist to be an aid to machakans and colossi, just as machakans exist to be an aid to pygmies and colossi. In other words, these are all kinda shite on their own but when mixed and matched (I plan to make better fitting themes/poses also) they will be much greater than the sum of their parts. I talked about this in Reddit Discord, but basically "Still I kind of want Colossi to be sort of cheaper and weaker than other size 3s, so that you dont want to bring them to a fight alone. For a good Austral nation, I am envisioning all three pop types each in a role. Like, Colossi nut crackers or machakan line holders or pygmy swarms as a frontline and then elites and then mages. Kinda shit if you go all the way with a single type, but together much stronger"

I am currently envisioning better pose roles and themes to be each theme having a specific role, be it elite/sacred, infantry, or mages. So with three types, that is 9 themes and then add in the tech level theme. The poses factor in by actually being the elite/sacred or infantry or mage.


Ealb:
The issue I have with this is that it essentially eliminates themes. The point of themes is that there CAN be Pheacia or Berytos - Colossi aren't just Machaka. These changes make more sense as themes for Colossi than as a blanket replacement whereby all Colossi are #primitive, no matter what the rest of the nation is.

PR #1184, opened 2021-01-17
Right Proper Austral Themeing

JAM-564:
Haha! I have used technomancy to create a "better" austral themeing thingy. Now, instead of predefined stuff, this'll make it so each  types get 3 place, troop/elite/mage. Then the austaul humas will generate and badabing, badaboom. Now you have a buncha ways to get australs, pygmy troops, machakan mages, colossi elites for one. It is wonderful and is only the beginning. The stuff it draws from is messed up to the nth power, stuff like machakan troops seeming to always generate with mounts and that kinda stuff. Anyway, it is here and I am happy.


elmokki:
This looks pretty cool. I might be tempted to make having other base machakans make more base machakans a bit more likely, but then again those are the least interesting nations.

PR #1186, opened 2021-01-18
"Balance" changes, aka: please let me play these they look so cool

JAM-564:
Most natgen nations are stronger to way stronger than vanilla. Even then, there are some natgen races that disregard every hope at being unbanned, and instead decide to be a meme. Crustaceans are basically unkillable unless you can consistently deal 30-35+ damage each swing, yet can kill things like no tomorrow. Cavernwights (Nadir) have buffs out the wazoo, from 15 MR to immunity to cold, to a chill aura.  So anyway, these things are getting gold cost inceases and crabs are getting a lower chance at 20 prot armour to combo with 10-15 natprot.

A notable exception is Dustwalkers, who will return to being 47 rec points instead of what they are now: 57.


Ealb:
Here's the thing about this: it's simply not true that "most NatGen nations are stronger to way stronger than vanilla" - it's that people dredge up and only use nations that are like that. If you randomly generate 40 nations, most are NOT stronger than vanilla by any stretch. NatGen nations are only wildly OP if you only play OP nations - the person who generates 10,000 nations to keep 1 will be more OP than someone who only generates 100. I know there's a certain degree of clash of playstyles here (I detest "bring your own nation NG" - the majority of NG games I've played were either draft from a pool with people getting to vote particular nations out of the pool, or full random), but trying to make NG cater to "generate 1000s and filter" setups runs counter to a lot of the design work put into it.

I also question boosting gold cost to try to impose balance. That tends to mean you can snowball, as gold is not a local resource like recpoints or resources. If you want crabs and cavernwights to be harder to recruit, make them harder to mass - and the best way to do that is to up their recpoints, not their gold cost. Even imposing an additional resource cost to represent the difficulty of re-animating them (or for crabs, upping their resource size to account for their bulky, complex carapaces and thus encouraging nudity) would do a better job of preventing them from massing than just making them cost more gold. The other thing with just leaning on gold costs is that it's not in line with vanilla.

elmokki:
While it's true most nationgen nations might not be way stronger than vanilla, some cool settings are indeed probably too powerful and should be balanced as much as possible without making them too weird. It's like the tengus were at some point when everyone picked them. 

I am more open to changing stats compared to base game than I have used to be. Especially for units that weren't originally meant to be recruitable.

Ealb:
I don't disagree - the point of things being added is to play them. However, I'm reluctant to lean too heavily into piecemail balance - I'd rather see the nation balanced than try to make individual units "balanced". That's generally how vanilla operates - some units would be OP if they were on other nations but aren't on the nations where they are available just because of the nation's other weaknesses.

elmokki:
Yes, ideally we would indeed force crappy mages and/or sacred for races with really strong troops. That said, ideally no single race would always be "yeah, strong roops, everything else must be bad", so I'm pretty sure there's room for unit-level balancing too.

Maybe we finally need to figure out what would make a good algorithm for figuring out if something is strong or weak and using that to balance nations!

JAM-564:
Most of the (extremely few) games of natgen I have played have used a sort of point system to add filters to the pack the organizer generates for you. In all of these crabs, nadir, and to a lesser extent dustwalkers have been banned or otherwise manually removed if they occur. 

On Crabs: When I was checking to see if these changes actually worked, I managed to accidently generate the pictured crab. Ignoring the regeneration and berserk, 26 hp and 28 prot is simply not something that can be dealt with by size 1 or 2 nations without using magic. And if they remain at such comparatively low gold costs, these plate armoured crabs (that are honestly quite common due to a lack of armor choices) will be somewhat of an extreme menace to deal with. Plus their innate second attack, 2 handed weapons, and size 3 makes them almost unstoppable in melee. At least Ulm has about human health pools, so that comparison that seemed apt to me is flawed. So yeah, that is my case that crabs should still be "nerfed". The method I agree is flawed, and that is why the gold increase is only on the elites and sacreds, and why the crabs had prot reduced from 15 to 10 on all but the elites and sacreds. And Pincer nerfed to not deal AP when not on elites and sacreds. But yeah, I still feel this is not enough and recognize that there has to be a better way to make the elites worse but still keep the spirit.
aaaaaaaaaaa

On Undead: The whole sentiment that undead should be harder to recruit and not greedier I strogly emphasizze with. Again, I could go on a spiel about how Dusties and Wights are overpowered even when randomly generated, but the purpose would be the same. Reiterate my point, pivot into trying out how to make them less banned, preserve the spirit, fit vanilla. The main undead nations seem to be freespawn, rely on non undead, or otherwise summon undead. My new idea is to try to get rid of recruitable wights and walkers and instead have them be custom summon spells. That would mess up withdustwalker/cavernwight only nation expansion, but I think that can easily be fixed with a few starting H1s (always reanimators and all) as well as keeping the undead in the starting army. That runs into the problem of trying to mess with summon costs, but I think if Loggy gets an interest (or I can bully the info out), his Magicgen thingy can be used as a baseline. 

I have mostly been talking about my proposed and committed changes in Discord, but if there is another place you think would be good for getting feedback, I would love to vet my stuff though more filters in the future.

Ealb:
Ha, I can't really say much about vetting changes since today was the first time I logged into Github or opened emails from it since September. I'm gonna try to re-engage with this, to include starting getting back on Discord. We'll see how well I can stick to that.

One thing that seems to be ignored a lot with crabs is how utterly awful their MR is. That makes me think of some other ways to weaken them besides just gutting their stats - they're simple-minded, so it's not unreasonable to pile onto that. While low morale doesn't fit with that, undisciplined does - as does increasing their space requirement w/o increasing their size a la Burning Ones and Lava Warriors.

I'm *very* reluctant to move dustwalker/cavernwight to summon-only - that's a nation-type we really don't have support for. I think they can be balanced. Scarcity is still the key to that, I'd say - giant nations have many of the same advantages but are still able to be dealt with. However, those also are manually balanced. Hmm. There are a few less violent hammers available to use before going to all-out removal, and not just scarcity - giants typically have uniform downsides beyond "they cost a lot". Adding random maluses to them is probably something worth considering.

geepope:
> Yes, ideally we would indeed force crappy mages and/or sacred for races with really strong troops. That said, ideally no single race would always be "yeah, strong roops, everything else must be bad", so I'm pretty sure there's room for unit-level balancing too.
> 
> Maybe we finally need to figure out what would make a good algorithm for figuring out if something is strong or weak and using that to balance nations!

When I had more free time to work on things I had started toying with the concept of algorithmically evaluating units, although I didn't get terribly far. My initial impulse was to go so far as to put together a simplified semi-deterministic Dom5 battle simulator using what mechanics are exposed, which has the advantage that it requires comparatively few value judgments. The drawback is that it became apparent that it would take a _lot_ of iterations to generate reliably accurate results, which could result in noticeably slower generation. A more probability-based approach to deriving average damage/TTK calculations would probably run faster, but would require more interpretation and judgment of results.

(Regardless, there are always going to be certain obstacles in trying to reduce troop quality into a single metric - if one unit outperforms another unit head-to-head, but is substantially less efficient at expanding, which troop is "better"? I suppose since the over-arching goal would be to balance nations holistically you could evaluate troops at a couple different roles so that you could take holes in a nation's capabilities into account.)

I do think that in the long run something along these lines is going to be necessary for balance, just because NationGen does not handle outliers well. In the community I play in, with the RC11 nerfs to crabs/dustwalkers I see them routinely eating dirt in games with relatively little cherry-picking (rolling ~3-5 nations per player, little to no filtering allowed)... but then you get outliers like the 28 prot berserk regen crab above. Balancing around the existence of monstrosities like that would nerf "average" crab nations into oblivion.

Drithyl:
Closing PR; cherrypicked the crabmen nerfs with some tweaks.
The Agarthan undead nerfs were unnecessary since they were already tweaked in previous changes.

PR #1189, opened 2021-03-26
Fix broken sacred slider, sacred balance pass

geepope:
Currently, the slider for sacreds' power level is broken and does not do anything; the only place in SacredGenerator that looks up the slider setting is in a method that does not actually use it. This moves the snippet to where it actually gets used to determine sacreds' power.


Techmind (approved):

Drithyl:
This has been addressed with a larger refactor of the sacredpower code and how sacred generation works, though some of the sacred filters and the survivability calculations have been pulled in and modified.

PR #1215, opened 2024-12-28
vanilla hoburgs are small have +3 def

Techmind:
Buffs hoburgs a bit having human level defence on 6 hp race is too bad,

### DIFF
--- a/data/races/hoburg.txt
+++ b/data/races/hoburg.txt
@@ -27,6 +27,7 @@
 #unitcommand "#enc 4"
 #unitcommand "#mor 9"
 #unitcommand "#twiceborn 3448"
+#unitcommand "#def 13"
 
 #nationcommand "#likespop 38"
 #nationcommand "#likespop 75"
--- a/data/races/hoburg.txt
+++ b/data/races/hoburg.txt
@@ -17,7 +17,6 @@
 #unitcommand "#size 2"
 #unitcommand "#hp 5"
 #unitcommand "#mr 14"
-#unitcommand "#def 12"
 #unitcommand "#str 6"
 #unitcommand "#prec 11"
 #unitcommand "#mapmove 12"
@@ -27,7 +26,7 @@
 #unitcommand "#enc 4"
 #unitcommand "#mor 9"
 #unitcommand "#twiceborn 3448"
-#unitcommand "#def 13"
+#unitcommand "#def +3"
 
 #nationcommand "#likespop 38"
 #nationcommand "#likespop 75"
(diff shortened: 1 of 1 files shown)


data/races/hoburg.txt
```diff
@@ -27,6 +27,7 @@
 #unitcommand "#enc 4"
 #unitcommand "#mor 9"
 #unitcommand "#twiceborn 3448"
+#unitcommand "#def 13"
```
lord-brimstone, line 30: `#def ` is up on line 20, change that instead

danieldelpercio: Yeah, I can approve it once that's been done. I checked hoburg stats in dominions 6 and it's a base of 13, so we can definitely bump this up.

Techmind: Changed to "+3", and removed double line , i think it works a bit better (based on limited testing i have done).

danieldelpercio: A bit better than what? Why does it work better. I thought having the base of 13 would be more in line with the formatting in the other templates.

Techmind: Well i saw that format used for adding glamour buffs or some other buffs and was thinking it might stack better with future changes, if you do it without "overwrite mode".

danieldelpercio: Seeing how this is the baseline for hoburgs I think it just makes it less clear and opaque, +3 from what? (The default 10, but it's easier to understand hoburgs are at 13 def if we have it defined that way in the hoburg file)

elmokki: Purely from the viewpoint of how NationGen has been made, these are the very first commands that are ever written for an unit. As such, a hard set here should not affect anything else from later on. Basesprite is the next one that's added, which is how some races with multiple very different subraces have been built.

danieldelpercio: I would prefer if line 20 was **#def 13**
- It's easier to figure out what the actual base value is without digging into code, easier for less technical users
- It's the same formatting as used in the other data/races files

elmokki: For future reference, the order commands are added to units starts with:

1. Race
2. Pose
3. Basesprite
4. Not sure exactly without digging deep into it, but it shouldn't matter

You should add statements like "#def 13" to the level that's the baseline for the thing you are doing. If one was doing basic Atlanteans and Shamblers, it'd make sense to differentiate the base stats in pose or basesprite. Iirc they're done in basesprite, but it should work at pose level too.

Techmind: Ok, rolling back to "def 13", makes sense if this is preferred format.

PR #1216, opened 2025-01-03
slider fix

Techmind:
Slider power fix, but without weird balance changes and with the same formatting.


Drithyl:
This has been addressed with a larger code refactoring of how sacredpower works.

PR #1230, opened 2025-03-08
bakemono cavalry fix

oldmidgetwillow:
Now they actually generate! Also some graphical fixes that required a katana pose.


Drithyl:
This has been merged into the danieldelpercio/natgen fork and will be merged to the main one through there.

